PR #2, opened 2018-03-27
Add pluginmanagement tag to pom.xml to resolve maven errors

aamanlamba:
The pluginManagement tag is required in pom.xml so it compiles fine - Eclipse is reporting errors on pom.xml without this tag

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -147,6 +147,7 @@
     </pluginRepositories>
 
     <build>
+    	<pluginManagement>
         <plugins>
             <plugin>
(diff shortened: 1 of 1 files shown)


jexp:
Thanks a lot, can you sign the Neo4j CLA? https://neo4j.com/developer/cla/
Otherwise I can also add it myself.
What's weird is that the current pom is valid, pluginManagement is just the optional "meta" handling.

aamanlamba:
Done

PR #52, opened 2018-10-29
Fix #30

### LINKED ISSUE #30: Provide a serialization format for Point Type

conker84:
We manage 2 different data structures for point type:

```
{"crs": "cartesian|cartesian-3d", "x": 1.1, "y": 13.4, "z": 3.1}
{"crs": "wgs-84|wgs-84-3d", "latitude": 13.1, "longitude": 13.4, "height": 1.1}
```

### DIFF
--- a/producer/src/main/kotlin/streams/serialization/JacksonUtil.kt
+++ b/producer/src/main/kotlin/streams/serialization/JacksonUtil.kt
@@ -2,20 +2,33 @@ package streams.serialization
 
 import org.codehaus.jackson.JsonGenerator
 import org.codehaus.jackson.JsonProcessingException
-import org.codehaus.jackson.map.ObjectMapper
 import org.codehaus.jackson.Version
 import org.codehaus.jackson.map.JsonSerializer
+import org.codehaus.jackson.map.ObjectMapper
 import org.codehaus.jackson.map.SerializationConfig
 import org.codehaus.jackson.map.SerializerProvider
 import org.codehaus.jackson.map.module.SimpleModule
-import org.neo4j.graphdb.spatial.CRS
-import org.neo4j.graphdb.spatial.Geometry
 import org.neo4j.graphdb.spatial.Point
+import org.neo4j.values.storable.CoordinateReferenceSystem
 import java.io.IOException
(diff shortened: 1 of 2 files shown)


producer/src/main/kotlin/streams/serialization/JacksonUtil.kt
```diff
@@ -2,28 +2,33 @@ package streams.serialization
...
+data class StreamsPointWgs(val crs: String, val latitude: Double, val longitude: Double, val height: Double?)
 class PointSerializer : JsonSerializer<Point>() {
     @Throws(IOException::class, JsonProcessingException::class)
     override fun serialize(value: Point?, jgen: JsonGenerator,
                            provider: SerializerProvider) {
         if (value == null) {
             return
         }
-        val point = InternalPoint(value!!.coordinates[0].coordinate, value!!.crs)
+        val crsType = value!!.crs.type
```
jexp: can we just add an extension method to InternalPoint?

jexp: So we can use it anywhere.

producer/src/main/kotlin/streams/serialization/JacksonUtil.kt
```diff
@@ -2,28 +2,33 @@ package streams.serialization
...
+        val coordinate = value!!.coordinates[0].coordinate
+        val point = if (crsType.startsWith(CoordinateReferenceSystem.Cartesian.toString())) {
```
jexp: shouldn't there be a more typesafe check?

jexp: shouldn't there be a more typesafe check?

producer/src/main/kotlin/streams/serialization/JacksonUtil.kt
jexp: can we just add an extension method to InternalPoint?
So we can use it anywhere.

conker84:
@jexp PR updated!

jexp:
Thanks a lot.

PR #72, opened 2018-11-09
Fix #42 - Performance test

### LINKED ISSUE #42: Performance Testing
Test with a real Kafka setup and inserting 1bn nodes and relationship into neo4j -> kafka
and also for the consumer 1bn events in kafka -> nodes and rels in neo4j

One nice real world test could be consuming the RSVP API endpoint for meetup.com
or the twitter streams API

we can also just have a simple (concurrent) create/update/delete workload that goes to kafka
and then on the other side uses the kafka sink to recreate the graph and then we compare the graphs that they are the same

omarlarus:
There's a new directory containing the python script that can be used also in a real use case to test the setup of the system. In the PR there's also the results of a local test as a chart.


performance/README.md
```diff
@@ -0,0 +1,135 @@
...
+- APOC 3.4.0.3
+- 2 GB heap memory
+- CREATE INDEX ON :Performance(group)
+- default neo4j-streams configuration
+- kafka.max.poll.records=16384
+
+The acquisition query:
+
+```properties
+streams.sink.topic.neo4j=WITH event.value.payload AS payload, event.value.meta AS meta CALL apoc.do.case( [\
```
jexp: this can be a regular "WHERE" filter :)

performance/README.md
```diff
@@ -0,0 +1,135 @@
...
+- 8 GB heap memory
+- CREATE INDEX ON :Performance(group)
+- default neo4j-streams configuration (without filtering)
+- kafka.batch.size=16384
+
+The  creation query used by the tester:
+
+	UNWIND range(1,{msg}) as ran
+	CREATE (n:Performance)
+	SET n.group = {uuid}, n.creation_time = apoc.date.currentTimestamp()
```
jexp: why is the uuid passed in from the outside?

performance/README.md
```diff
@@ -0,0 +1,135 @@
...
+Currently, it's available only the test of neo4j to neo4j configuration, in remote too.
+
+## Setup
+
+The basic unit of work used to test the performances needs a specific cypher part in order to get the timestamp of creation and receiving of the event. The elapsed time is calculated using these timestamps through cypher queries.
+
+### Producer
+
+- neo4j 3.4.7 community edition
+- APOC 3.4.0.3
```
jexp: which kafka version?
we probably shouldn't need apoc.

consumer/src/main/kotlin/streams/StreamsEventSinkQueryExecution.kt
```diff
@@ -12,6 +12,7 @@ class StreamsEventSinkQueryExecution(private val streamsTopicService: StreamsTop
         if (cypherQuery == null) {
             return
         }
+        log.info("UNWIND ${params.size} events from Kafka")
```
jexp: shouldn't this be a debug output?
with if (log.isDebugEnabled) around it.

jexp: change the message to "Processing ...."

performance/README.md
```diff
@@ -0,0 +1,135 @@
...
+
+To run the test you need to edit the *config.ini* file in order to setup the connection data for the *producer* and *consumer* neo4j instances.
+
+The *unit* part is used as default values to run the tests. 
+
+- *repeat* is the number of times each test is executed
+- *node* is the number of nodes created for each test
+
+```ini
+[producer]
```
jexp: shouldn't this be handled by the docker compose?

jexp:
Do we have a comparison of using the producer vs. not?

jexp:
It would be good if this was more repeatable:

i.e. using docker compose with
- 2x neo4j
- 1x kafka + zk
- 1x tester with python 3.6 and the test script writing it's outputs (csv) to a host directory
- using compose links to make the services visible to each other

- somehow the python bits look a bit involved
- basically it's just an repeat of fixed operations
- operation x create k nodes in a tx
- where the operation x could end up as property on the node instead of the uuid
- and then the output would just be a single summary cypher statement of operations, min,max,stddev of time-delta time per operation

performance/neo4j-streams-pt.py
```diff
@@ -0,0 +1,107 @@
...
+			MATCH (n:Performance)
+			WHERE n.group = {uuid}
+			RETURN count(n) as cnt, max(n.received_time) as received_time
+			""", uuid=test_ref).data()
+
+		#print(result)
+		in_streaming = result[0]['cnt'] < nodes
+		received_time = result[0]['received_time']
+
+	return (received_time - creation_time) / nodes
```
jexp: probably better to plot it not as percentage but use nodes as X axis and then the  time delta

performance/README.md
omarlarus: in real scenario, apoc.do.case may be more useful to split the cases

performance/README.md
omarlarus: because the uuid it's only a way to identify in the cunsumer the group of records already created in the producer, generated by the code and passed into the query. The query never changes.

performance/README.md
omarlarus: Updated the PR with the kafka version. The apoc is used to put the timestamp into the nodes created. Using the apoc.date.currentTimestamp() each node has a different timestamp in the same transaction, so I can pick the min in the producer and the max in the consumer

consumer/src/main/kotlin/streams/StreamsEventSinkQueryExecution.kt
omarlarus: Updated PR

performance/README.md
omarlarus: A config file is useful when someone wants to test a real environment.

performance/neo4j-streams-pt.py
omarlarus: It's easier to compare the milliseconds per node instead of the whole elapsed time between very different series (ms vs sec or minutes).

PR #79, opened 2018-11-12
fixes #70: Document Docker usage/configuration in readme/docs

### LINKED ISSUE #70: Document Docker usage/configuration in readme/docs
Especially the `.` to underscore treatment and where to put the jar.
And underscore to double underscore.

So for folks that run Neo4j in docker they can use this.

conker84:


doc/asciidoc/docker/index.adoc
```diff
@@ -0,0 +1,112 @@
...
+{"key":"neo4j","value":{"meta":{"timestamp":1542047038549,"username":"neo4j","txId":12,"txEventId":108,"txEventsCount":110,"operation":"created","source":{"hostname":"neo4j"}},"payload":{"id":"96","start":{"id":"9","labels":["Person"]},"end":{"id":"7","labels":["Person"]},"before":null,"after":{"properties":{"years":2}},"label":"KNOWS","type":"relationship"},"schema":{"properties":[],"constraints":null}}}
+{"key":"neo4j","value":{"meta":{"timestamp":1542047038549,"username":"neo4j","txId":12,"txEventId":109,"txEventsCount":110,"operation":"created","source":{"hostname":"neo4j"}},"payload":{"id":"97","start":{"id":"9","labels":["Person"]},"end":{"id":"8","labels":["Person"]},"before":null,"after":{"properties":{"years":1}},"label":"KNOWS","type":"relationship"},"schema":{"properties":[],"constraints":null}}}
+```
+
+=== Consumer
+
+In case of you are using the Sink you can define your topic/cypher-query combination as it follows:
+
+```yml
```
jexp: We definitely need to simplify this in the future.
Usually, I guess the topics would be separated by type.
and then we could do just do one query by topic and use: `WITH * WHERE meta.operation <> 'deleted'`

doc/asciidoc/docker/data/docker-compose.yml
```diff
@@ -0,0 +1,41 @@
...
+    image: neo4j:3.4
+    hostname: neo4j
+    container_name: neo4j
+    ports:
+    - "7474:7474"
+    - "7687:7687"
+    depends_on:
+    - kafka
+    volumes:
+    - $HOME/neo4j/3.4/plugins:/plugins
```
jexp: I guess this is something that this then is externalized. Perhaps we can put a comment into the YML.

PR #85, opened 2018-11-23
fixes #84: Add a stream.publish(topic, payload) procedure

### LINKED ISSUE #84: Add a stream.publish(topic, payload) procedure
To allow custom streaming from Neo4j to the configured stream environment.

topic is a string, payload is just a map, there should be an optional config as third option.

payload could also be a node or relationship then we should use the same converter as for the current producer
also we need ot handle nested values that might contain nodes, rels, lists or points/datetime.

conker84:
Fixes #84 

Added a procedure `streams.publish` that allows sending custom messages

 - Added Lifecycle methods to `StreamsEventRouter`
 - Changed **StreamsEvent** to `open class StreamsEvent(open val payload: Any)` and created a new class **StreamsTransactionEvent** as `data class StreamsTransactionEvent(val meta: Meta, override val payload: Payload, val schema: Schema): StreamsEvent(payload)` that manages the transaction event event
 - Changed the method `StreamsEventRouter##sendEvents` in order to get topic as input and use it from the `streams.publish` procedure
 - The procedure Manages all kind of Neo4jTypes, in case of Node/Relationship if the topic is has filtering rules all the properties will be filtered according to those rules
 - Updated the documentation

### DIFF
--- a/consumer/src/main/kotlin/streams/StreamsEventSinkQueryExecution.kt
+++ b/consumer/src/main/kotlin/streams/StreamsEventSinkQueryExecution.kt
@@ -13,7 +13,6 @@ class StreamsEventSinkQueryExecution(private val streamsTopicService: StreamsTop
             return
         }
         if(log.isDebugEnabled){
-
             log.debug("Processing ${params.size} events from Kafka")
         }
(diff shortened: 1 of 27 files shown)


jexp:
Thanks soo much !!

Issue #51, opened 2018-10-27
Parametrize Integration tests


conker84:
In order to test the system with different Kafka versions

PR #101, opened 2018-12-10
fixes #96: New procedure to receive data from the stream

### LINKED ISSUE #96: New procedure to receive data from the stream
```
streams.receive(topic, {timeout:ms, commit:from-beginning/-1, events:100, ...}) yield event
```

conker84:
Fixes #96 
Created a procedure `streams.consume` that allows to create consume custom topics at runtime. Example of usage:

```
CALL streams.consume('my-topic', {<config>}) YIELD event
CREATE (p:Person{firstName: event.name, lastName: event.surname})
```

Kafka does not allow to define the number of event that a user want to collect as requested: `events:100`, it has a property `fetch.min.bytes` that _is the minimum amount of data the server should return for a fetch request_ (currently this property is not mapped, let me know if you want it)

  - refactored the Sink in order to expose a common apis that allows to create custom cosumer
  - created the procedure `streams.consume`
  - updated the documentation

### DIFF
--- a/common/src/main/kotlin/streams/utils/StreamsUtils.kt
+++ b/common/src/main/kotlin/streams/utils/StreamsUtils.kt
@@ -12,7 +12,10 @@ object StreamsUtils {
         return try {
             action()
         } catch (e: Throwable) {
-            when (e::class.java) {
+            if (toIgnore.isEmpty()) {
+                return null
+            }
+            return when (e::class.java) {
                 in toIgnore -> null
                 else -> throw e
             }
--- a/consumer/src/main/kotlin/streams/StreamsEventSinkQueryExecution.kt
+++ b/consumer/src/main/kotlin/streams/StreamsEventSinkQueryExecution.kt
@@ -1,21 +1,34 @@
 package streams
 
+import org.neo4j.graphdb.GraphDatabaseService
 import org.neo4j.kernel.internal.GraphDatabaseAPI
 import org.neo4j.logging.Log
+import streams.utils.Neo4jUtils
+import streams.utils.StreamsUtils
 
 class StreamsEventSinkQueryExecution(private val streamsTopicService: StreamsTopicService, private val db: GraphDatabaseAPI, val log: Log) {
-
-    private val UNWIND: String = "UNWIND {events} AS event"
 
     fun execute(topic: String, params: Collection<Any>) {
         val cypherQuery = streamsTopicService.get(topic)
         if (cypherQuery == null) {
             return
         }
+        val query = "${StreamsUtils.UNWIND} $cypherQuery"
         if(log.isDebugEnabled){
-            log.debug("Processing ${params.size} events from Kafka")
+            log.debug("Processing ${params.size} events with query: $query")
         }
-        db.execute("$UNWIND $cypherQuery", mapOf("events" to params)).close()
+        if (Neo4jUtils.isWriteableInstance(db)) {
+            try {
+                db.execute(query, mapOf("events" to params)).close()
+            } catch (e: Exception) {
+                log.error("Error while executing the query", e)
+            }
+        } else {
+            if(log.isDebugEnabled){
+                log.debug("Not writeable instance")
+            }
+        }
+
     }
 
     fun execute(map: Map<String, Collection<Any>>) {
--- a/consumer/src/main/kotlin/streams/StreamsTopicService.kt
+++ b/consumer/src/main/kotlin/streams/StreamsTopicService.kt
@@ -3,19 +3,45 @@ package streams
 import org.apache.commons.lang3.StringUtils
 import org.neo4j.kernel.impl.core.EmbeddedProxySPI
 import org.neo4j.kernel.impl.core.GraphProperties
-import org.neo4j.kernel.impl.logging.LogService
 import org.neo4j.kernel.internal.GraphDatabaseAPI
 import streams.utils.Neo4jUtils
 
 
 private const val STREAMS_TOPIC_KEY: String = "streams.sink.topic."
 
-class StreamsTopicService(private val db: GraphDatabaseAPI, private val streamsSinkConfiguration: StreamsSinkConfiguration) {
+class StreamsTopicService(private val db: GraphDatabaseAPI, private val topicMap: Map<String, String>) {
     private val properties: GraphProperties = db.dependencyResolver.resolveDependency(EmbeddedProxySPI::class.java).newGraphPropertiesProxy()
     private val log = Neo4jUtils.getLogService(db).getUserLog(StreamsTopicService::class.java)
 
     init {
-        setAll(streamsSinkConfiguration.topics)
+        clear()
+        setAll(topicMap)
+    }
+
+    fun clear() {
+        return db.beginTx().use {
+            val keys = properties.allProperties
+                    .filterKeys { it.startsWith(STREAMS_TOPIC_KEY) }
+                    .keys
+            keys.forEach {
+                properties.removeProperty(it)
+            }
+            it.success()
+        }
+    }
+
+    fun remove(topic: String) {
+        val key = "$STREAMS_TOPIC_KEY$topic"
+        return db.beginTx().use {
+            if (!properties.hasProperty(key)) {
+                if (log.isDebugEnabled) {
+                    log.debug("No query registered for topic $topic")
+                }
+                return
+            }
+            properties.removeProperty(key)
+            it.success()
+        }
     }
 
     fun set(topic: String, query: String) {
--- a/consumer/src/main/kotlin/streams/kafka/KafkaEventSink.kt
+++ b/consumer/src/main/kotlin/streams/kafka/KafkaEventSink.kt
@@ -4,109 +4,133 @@ import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.GlobalScope
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.launch
-import org.apache.kafka.clients.consumer.ConsumerRecords
 import org.apache.kafka.clients.consumer.KafkaConsumer
 import org.apache.kafka.common.errors.WakeupException
-import org.codehaus.jackson.map.ObjectMapper
 import org.neo4j.kernel.configuration.Config
-import org.neo4j.kernel.internal.GraphDatabaseAPI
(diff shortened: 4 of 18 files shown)


jexp:
What about managing the commit retrieval via the config?
like start at beginning or start at tx xxx ?

common/src/main/kotlin/streams/utils/StreamsUtils.kt
```diff
@@ -12,7 +12,10 @@ object StreamsUtils {
         return try {
             action()
         } catch (e: Throwable) {
-            when (e::class.java) {
+            if (toIgnore.isEmpty()) {
```
jexp: doesn't toIgnore empty mean that you shouldn't ignore any?

consumer/src/main/kotlin/streams/procedures/StreamsSinkProcedures.kt
```diff
@@ -0,0 +1,73 @@
...
+    @Description("streams.consume(topic, config) - Allows to subscribe custom topics")
+    fun consume(@Name("topic") topic: String?,
+                @Name(value = "config", defaultValue = "{}") config: Map<String, Any>?): Stream<StreamResult> {
+        checkEnabled()
+        if (topic == null || topic == StringUtils.EMPTY) {
+            log?.info("Topic empty, no message sent")
+            return Stream.empty()
+        }
+
+        val properties = if (config == null) {
```
jexp: you can use `config?.mapValues { it.value.toString() } ?: emptyMap()`

consumer/src/main/kotlin/streams/procedures/StreamsSinkProcedures.kt
```diff
@@ -0,0 +1,73 @@
...
+
+    @JvmField @Context
+    var log: Log? = null
+
+    @Procedure(mode = Mode.SCHEMA, name = "streams.consume")
+    @Description("streams.consume(topic, config) - Allows to subscribe custom topics")
+    fun consume(@Name("topic") topic: String?,
+                @Name(value = "config", defaultValue = "{}") config: Map<String, Any>?): Stream<StreamResult> {
+        checkEnabled()
+        if (topic == null || topic == StringUtils.EMPTY) {
```
jexp: `topic.isNullOrEmpty()`

consumer/src/main/kotlin/streams/procedures/StreamsSinkProcedures.kt
```diff
@@ -0,0 +1,73 @@
...
+import streams.StreamsEventSinkConfigMapper
+import streams.StreamsSinkConfiguration
+import java.util.stream.Stream
+
+class StreamResult(@JvmField val event: Any)
+
+class StreamsSinkProcedures {
+
+    @JvmField @Context
+    var log: Log? = null
```
jexp: can we make this lateinit?

consumer/src/main/kotlin/streams/StreamsTopicService.kt
```diff
@@ -3,19 +3,45 @@ package streams
...
     init {
-        setAll(streamsSinkConfiguration.topics)
+        clear()
+        setAll(topicMap)
+    }
+
+    fun clear() {
+        return db.beginTx().use {
+            val keys = properties.allProperties
+                    .filter { it.key.startsWith(STREAMS_TOPIC_KEY) }
```
jexp: keys.filter

consumer/src/main/kotlin/streams/kafka/KafkaEventSink.kt
```diff
@@ -4,109 +4,133 @@ import kotlinx.coroutines.Dispatchers
...
-        if (log.isDebugEnabled) {
-            log.debug("Subscribing topics: ${streamsTopicService!!.getAll()}")
-        } else {
-            log.info("Subscribing topics: ${streamsTopicService!!.getTopics()}")
+    override fun getEventSinkConfigMapper(): StreamsEventSinkConfigMapper {
+        return object: StreamsEventSinkConfigMapper(streamsConfigMap, mappingKeys) {
+            override fun convert(config: Map<String, String>): Map<String, String> {
+                val props = streamsConfigMap
+                        .toMutableMap()
+                props.putAll(config.mapKeys { mappingKeys.getOrDefault(it.key, it.key) })
```
jexp: what does this do again?
btw. youc an just do `= streamsConfigMap + config.mapKeys { mappingKeys.getOrDefault(it.key, it.key) }`
no need for the mutable map.

consumer/src/main/kotlin/streams/kafka/KafkaEventSink.kt
```diff
@@ -4,109 +4,133 @@ import kotlinx.coroutines.Dispatchers
...
+                        }
+                        queryExecution.execute(it.key, it.value)
                     }
                 }
             } catch (e: Throwable) {
                 val message = e.message ?: "Generic error, please check the stack trace: "
                 log.error(message, e)
-            } finally {
-                kafkaConsumer.close()
+                eventConsumer.stop()
```
jexp: that means at. the first error it stops the consumer?

consumer/src/test/kotlin/integrations/StreamsSinkProceduresIT.kt
```diff
@@ -0,0 +1,126 @@
...
+         *
+         * Confluent Platform | Apache Kafka
+         *                    |
+         * 4.0.x	          | 1.0.x
+         * 4.1.x	          | 1.1.x
+         * 5.0.x	          | 2.0.x
+         *
+         * Please see also https://docs.confluent.io/current/installation/versions-interoperability.html#cp-and-apache-kafka-compatibility
+         */
+        private const val confluentPlatformVersion = "4.0.2"
```
jexp: are we planning to test the other versions too?
or is that not needed?

jexp: perhaps we can do it with a parameterized test or subclasses?

consumer/src/main/kotlin/streams/StreamsEventSinkQueryExecution.kt
```diff
@@ -1,21 +1,21 @@
...
+import org.neo4j.graphdb.GraphDatabaseService
 import org.neo4j.logging.Log
+import streams.utils.StreamsUtils
 
-class StreamsEventSinkQueryExecution(private val streamsTopicService: StreamsTopicService, private val db: GraphDatabaseAPI, val log: Log) {
-
-    private val UNWIND: String = "UNWIND {events} AS event"
+class StreamsEventSinkQueryExecution(private val streamsTopicService: StreamsTopicService, private val db: GraphDatabaseService, val log: Log) {
 
     fun execute(topic: String, params: Collection<Any>) {
```
jexp: this doesn't handle any exceptions so the consumer will abort it's data loop on the first cypher error?
we should handle that better, e.g. look at transient vs. non-transient errors
and either just send the message back to Kafka (i.e. don't commit reception) for retry
or disable the topic statement after a number of errors and log that.

jexp: possibly also log error messages here.

jexp: read vs. write permissions might be ignored though as this can happen in a cluster by intent.

common/src/main/kotlin/streams/utils/StreamsUtils.kt
conker84: I was thinking that means "ignore all" because otherwise I simply do not use it. But I can revert it!

consumer/src/main/kotlin/streams/procedures/StreamsSinkProcedures.kt
conker84: Ok!

consumer/src/main/kotlin/streams/procedures/StreamsSinkProcedures.kt
conker84: Ok!

consumer/src/main/kotlin/streams/kafka/KafkaEventSink.kt
conker84: Ok!

consumer/src/test/kotlin/integrations/StreamsSinkProceduresIT.kt
conker84: We have the #51 that covers it

common/src/main/kotlin/streams/utils/StreamsUtils.kt
jexp: It's fine.

consumer/src/main/kotlin/streams/kafka/KafkaEventSink.kt
jexp: and then there is no way to restart it without restarting the server?

consumer/src/main/kotlin/streams/procedures/StreamsSinkProcedures.kt
conker84: Unfortunately `@JvmField` (which is required to expose de field as _Java public_) is incompatible with `lateinit`

consumer/src/main/kotlin/streams/kafka/KafkaEventSink.kt
conker84: It merges the default configuration such as zookeeper and broker addresses provided in the Neo4j conf with the custom that comes from the `{config}` param of the procedure

consumer/src/main/kotlin/streams/StreamsTopicService.kt
conker84: Ok!

conker84:
@jexp PR updated!

conker84:
> What about managing the commit retrieval via the config?
> like start at beginning or start at tx xxx ?

That is something that we can do but we have to change the following auto-commit behaviour to a manual management like spark do:

https://github.com/apache/spark/blob/5264164a67df498b73facae207eda12ee133be7d/external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/KafkaOffsetReader.scala#L132-L158

PR #133, opened 2019-01-15
Kafka-Connect improvements 

conker84:

  - added config property `neo4j.batch.size` that splits the records into chunks with the provided value
  - changed `pom.xml` in order to simplify the build process
  - updated `docker-compose.yml` in order to support *neo4j-enterprise*


kafka-connect-neo4j/src/main/kotlin/streams/kafka/connect/sink/Neo4jSinkTask.kt
```diff
@@ -37,16 +33,31 @@ class Neo4jSinkTask : SinkTask() {
...
                 }
                 .mapKeys { it -> "${StreamsUtils.UNWIND} ${config.topicMap[it.key]}" }
-                .mapValues { it -> mapOf("events" to it.value.map { converter.convert(it.value()) }) }
+                .mapValues { it ->
+                    val value = it.value
+                    val chunks = value.chunked(config.batchSize)
+                    if (log.isDebugEnabled) {
+                        log.debug("Data chunked in ${chunks.size} chunks")
+                    }
+                    val result = chunks.map {
```
jexp: does converter convert handle collections?
also if I understand correctly we now changed key -> list<value> to key -< list<list<value>> ?
bu then the data processing didn't change?

jexp:
Also I didn't see a test for the batching?

kafka-connect-neo4j/src/main/kotlin/streams/kafka/connect/sink/Neo4jSinkTask.kt
conker84: - No, I searched through different connectors [here an example](https://github.com/hpgrahsl/kafka-connect-mongodb/blob/master/src/main/java/at/grahsl/kafka/connect/mongodb/converter/SinkConverter.java) and the all manage `Struct` and `Map`. Moreover our `converter` extends an utility class called [AbstractConverter](https://github.com/jcustenborder/connect-utils/blob/3d2db83577825e6cd479fa14da2919521a6cd75d/connect-utils/src/main/java/com/github/jcustenborder/kafka/connect/utils/data/AbstractConverter.java#L73) create by a guy that works in Confluent. Maybe this is something that we can go deepen with the Confluent guys;
- yes we use the new property `config.batchSize` in order to split the data in chunks of the required size; generally speaking we follow these steps (starting from a `Collection<SinkRecords>`):
  - we group the data by the topic;
  - we filter the data by valid topics;
  - we transform the keys to the corrisponding templated queries;
  - for each topic we transform the values by chunking the dataset;
this brings us to what you are saying;
yes, now we have the < key, list< list >> and for now we simply split the source dataset into chunks, but we can provide a more elaborated behavior (like a queue that accumulates sink records and dequeue the data based on chunk size or timeout) in the near future.

PR #135, opened 2019-01-17
follow-up: Kafka-Connect improvements

conker84:

  - created the `EventBuilder` class that manages all the business logic for chunking data; created the test case `EventBuilderTest`;
  - fixed list of list bug

### DIFF
--- a/kafka-connect-neo4j/src/main/kotlin/streams/kafka/connect/sink/EventBuilder.kt
+++ b/kafka-connect-neo4j/src/main/kotlin/streams/kafka/connect/sink/EventBuilder.kt
@@ -1,0 +1,45 @@
+package streams.kafka.connect.sink
+
+import org.apache.kafka.connect.sink.SinkRecord
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+
+class EventBuilder {
+    private val log: Logger = LoggerFactory.getLogger(EventBuilder::class.java)
+
+    private var batchSize: Int? = null
+    private lateinit var topics: Set<String>
+    private lateinit var sinkRecords: Collection<SinkRecord>
+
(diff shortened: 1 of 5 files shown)


kafka-connect-neo4j/src/main/kotlin/streams/kafka/connect/sink/EventBuilder.kt
```diff
@@ -0,0 +1,45 @@
...
+    fun withSinkRecords(sinkRecords: Collection<SinkRecord>): EventBuilder {
+        this.sinkRecords = sinkRecords
+        return this
+    }
+
+    fun build(): Map<String, List<List<SinkRecord>>> { // <Topic, List<List<SinkRecord>>
+        return this.sinkRecords
+                .groupBy { it.topic() }
+                .filterKeys {topic ->
+                    val isValidTopic = topics.contains(topic)
```
jexp: perhaps we should filter before the group by?

kafka-connect-neo4j/src/main/kotlin/streams/kafka/connect/sink/EventBuilder.kt
```diff
@@ -0,0 +1,45 @@
...
+import org.slf4j.LoggerFactory
+
+class EventBuilder {
+    private val log: Logger = LoggerFactory.getLogger(EventBuilder::class.java)
+
+    private var batchSize: Int? = null
+    private lateinit var topics: Set<String>
+    private lateinit var sinkRecords: Collection<SinkRecord>
+
+    fun withBatchSize(batchSize: Int): EventBuilder {
```
jexp: do we need the builder pattern here?
Why not just a class with batchsize and topics 
and then apply/prepare or similar method that takes the records

PR #161, opened 2019-03-11
fixes issue #156; #158

### LINKED ISSUE #156: make streams.publish return its arguments when it is successful
void procedures are hard to chain together with other stuff in cypher.
Thanks to @moxious

### LINKED ISSUE #158: streams.produce doesn't publish in case of null argument
CALL streams.publish(‘test’, null) does not publish { payload: null }.
Provide an exception in order to highlight to the user the behaviour.

moxious:
Fixes #156 
Fixes #158 

CALL streams.publish(a, b, c) returns its arguments when it is successful, allowing verification that the message was published, and chaining of arguments in cypher.

### DIFF
--- a/producer/src/main/kotlin/streams/kafka/KafkaEventRouter.kt
+++ b/producer/src/main/kotlin/streams/kafka/KafkaEventRouter.kt
@@ -16,6 +16,7 @@ import streams.events.StreamsEvent
 import streams.events.StreamsTransactionEvent
 import streams.serialization.JSONUtils
 import java.util.*
+import java.util.concurrent.Future
 import java.util.concurrent.ThreadLocalRandom
 
 
@@ -42,43 +43,44 @@ class KafkaEventRouter: StreamsEventRouter {
         producer.close()
     }
 
-    private fun send(producerRecord: ProducerRecord<String, ByteArray>) {
-        producer.send(producerRecord) { meta: RecordMetadata?, error: Exception? ->
+    private fun send(producerRecord: ProducerRecord<String, ByteArray>): Future<RecordMetadata> {
+        return producer.send(producerRecord) { meta: RecordMetadata?, error: Exception? ->
             if (log.isDebugEnabled) {
                 log.debug("Sent record in partition ${meta?.partition()} offset ${meta?.offset()} data ${meta?.topic()} key size ${meta?.serializedKeySize()}", error)
             }
         }
     }
 
-    private fun sendEvent(partition: Int, topic: String, event: StreamsEvent) {
+    private fun sendEvent(partition: Int, topic: String, event: StreamsEvent): Future<RecordMetadata> {
         if (log.isDebugEnabled) {
             log.debug("Trying to send a simple event with payload ${event.payload} to kafka")
         }
         val uuid = UUID.randomUUID().toString()
         val producerRecord = ProducerRecord(topic, partition, System.currentTimeMillis(), uuid,
                 JSONUtils.writeValueAsBytes(event))
-        send(producerRecord)
+        return send(producerRecord)
     }
 
-    private fun sendEvent(partition: Int, topic: String, event: StreamsTransactionEvent) {
+    private fun sendEvent(partition: Int, topic: String, event: StreamsTransactionEvent): Future<RecordMetadata> {
         if (log.isDebugEnabled) {
             log.debug("Trying to send a transaction event with txId ${event.meta.txId} and txEventId ${event.meta.txEventId} to kafka")
         }
         val producerRecord = ProducerRecord(topic, partition, System.currentTimeMillis(), "${event.meta.txId + event.meta.txEventId}-${event.meta.txEventId}",
                 JSONUtils.writeValueAsBytes(event))
-        send(producerRecord)
+        return send(producerRecord)
     }
 
-    override fun sendEvents(topic: String, transactionEvents: List<out StreamsEvent>) {
-        try {
+    override fun sendEvents(topic: String, transactionEvents: List<out StreamsEvent>): List<out Future<RecordMetadata>> {
+        val sendFutures: MutableList<Future<RecordMetadata>> = mutableListOf()
+
+         try {
             producer.beginTransaction()
             transactionEvents.forEach {
+                // Use a mutable list and add each one, because we might send some and get an exception
+                // provoked by later ones.  We want to be able to return all sendFutures we managed to
+                // construct before hitting the Exception.
                 val partition = ThreadLocalRandom.current().nextInt(kafkaConfig.numPartitions)
-                if (it is StreamsTransactionEvent) {
-                    sendEvent(partition, topic, it)
-                } else {
-                    sendEvent(partition, topic, it)
-                }
+        sendFutures.add(sendEvent(partition, topic, it))
             }
             producer.commitTransaction()
         } catch (e: ProducerFencedException) {
@@ -94,8 +96,19 @@ class KafkaEventRouter: StreamsEventRouter {
             log.error("Generic kafka error. Stack trace is:", e)
             producer.abortTransaction()
         }
+
+        return sendFutures
     }
 
+    override fun sendEventsSync(topic: String, transactionEvents: List<out StreamsEvent>): List<out RecordMetadata> {
+        producer.beginTransaction();
+
+        var results = transactionEvents.map { it ->
+            sendEvent(ThreadLocalRandom.current().nextInt(kafkaConfig.numPartitions), topic, it) }
+                .map { it -> it.get() }
+        producer.commitTransaction()
+        return results
+    }
 }
 
 class Neo4jKafkaProducer<K, V>: KafkaProducer<K, V> {
--- a/producer/src/main/kotlin/streams/procedures/StreamsProcedures.kt
+++ b/producer/src/main/kotlin/streams/procedures/StreamsProcedures.kt
@@ -7,6 +7,18 @@ import streams.StreamsEventRouter
 import streams.StreamsEventRouterConfiguration
 import streams.events.StreamsEventBuilder
 import java.lang.RuntimeException
+import java.util.stream.Stream
+
+class PublishResult {
+    @JvmField public var topic: String
+    @JvmField public var payload: Any
+    @JvmField public var config: Map<String,Any>?
+    constructor(topic: String, payload: Any, config: Map<String,Any>?) {
+        this.topic = topic;
+        this.payload = payload;
+        this.config = config;
+    }
+}
 
 class StreamsProcedures {
 
@@ -15,15 +27,15 @@ class StreamsProcedures {
     @Procedure(mode = Mode.SCHEMA, name = "streams.publish")
     @Description("streams.publish(topic, config) - Allows custom streaming from Neo4j to the configured stream environment")
     fun publish(@Name("topic") topic: String?, @Name("payload") payload: Any?,
-                @Name(value = "config", defaultValue = "{}") config: Map<String, Any>?) {
+                @Name(value = "config", defaultValue = "{}") config: Map<String, Any>?): Stream<PublishResult> {
         checkEnabled()
         if (topic.isNullOrEmpty()) {
             log?.info("Topic empty, no message sent")
-            return
+            return Stream.empty();
         }
         if (payload == null) {
             log?.info("Payload empty, no message sent")
-            return
+            return Stream.empty();
         }
         val streamsEvent = StreamsEventBuilder()
                 .withPayload(payload)
@@ -38,6 +50,8 @@ class StreamsProcedures {
                 .withTopic(topic)
                 .build()
         StreamsProcedures.eventRouter.sendEvents(topic, listOf(streamsEvent))
+
+        return Stream.of(PublishResult(topic, payload, config))
     }
 
     private fun checkEnabled() {
--- a/producer/src/main/kotlin/streams/procedures/StreamsProcedures.kt
+++ b/producer/src/main/kotlin/streams/procedures/StreamsProcedures.kt
@@ -1,6 +1,7 @@
 package streams.procedures
 
 import org.apache.commons.lang3.StringUtils
+import org.apache.kafka.clients.producer.RecordMetadata
 import org.neo4j.logging.Log
 import org.neo4j.procedure.*
 import streams.StreamsEventRouter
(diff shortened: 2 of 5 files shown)


moxious:
Achievement unlocked -- my first neo4j-streams PR and my first ever written line of Kotlin.  :)

producer/src/main/kotlin/streams/procedures/StreamsProcedures.kt
```diff
@@ -15,15 +27,15 @@ class StreamsProcedures {
     @Procedure(mode = Mode.SCHEMA, name = "streams.publish")
     @Description("streams.publish(topic, config) - Allows custom streaming from Neo4j to the configured stream environment")
     fun publish(@Name("topic") topic: String?, @Name("payload") payload: Any?,
-                @Name(value = "config", defaultValue = "{}") config: Map<String, Any>?) {
+                @Name(value = "config", defaultValue = "{}") config: Map<String, Any>?): Stream<PublishResult> {
         checkEnabled()
         if (topic.isNullOrEmpty()) {
             log?.info("Topic empty, no message sent")
```
jexp: should we rather return this to the client

producer/src/main/kotlin/streams/procedures/StreamsProcedures.kt
```diff
@@ -38,6 +50,8 @@ class StreamsProcedures {
                 .withTopic(topic)
                 .build()
         StreamsProcedures.eventRouter.sendEvents(topic, listOf(streamsEvent))
```
jexp: this is asynchronous, afaik 

so we're not sure if the event was actually sent, it's more like fire&forget.

producer/src/test/kotlin/streams/integrations/KafkaEventRouterIT.kt
```diff
@@ -154,7 +156,15 @@ class KafkaEventRouterIT {
...
         val message = "Hello World"
-        db.execute("CALL streams.publish('neo4j', '$message')").close()
+
+        val result = db.execute("CALL streams.publish('neo4j', '$message')")
+        assertTrue { result.hasNext() }
+        val resultMap = result.next()
+
+        assertTrue { "$message".equals(resultMap.get("payload")) }
+        assertTrue { "neo4j".equals(resultMap.get("topic")) }
+
```
jexp: assertFalse result.hasNext

producer/src/main/kotlin/streams/procedures/StreamsProcedures.kt
jexp: @conker84 wdyt

conker84: Yes it's an async call. 
It would be cool to add a new method `sendSync` into StreamsEventRouter and implement it in the `KafkaEventRouter` in this way:

```kotlin
    override fun startSync(producerRecord: ProducerRecord<String, ByteArray>): RecordMetadata {
        producer.send(producerRecord).get()
    }
``` 

So we can provide as output the `RecordMetadata`

moxious: ah you're right this is even better.  I'll look into it.

producer/src/main/kotlin/streams/procedures/StreamsProcedures.kt
moxious: I don't understand this comment.  Do you mean we should not return config, or do you mean we should find a way to return the log message if topic is empty?

jexp: I meant that no one will see the log. And if we're already returning stuff we should probably return this info (that it was actually not sent instead of an empty stream).

moxious: You are right -- this is existing code and I would like to keep it to a different PR some other time though for several reasons -- 

When you are sending multiple, it's ambiguous whether you mean "best effort" deliver as many as you can, or if you mean fail if one error happens.  Also if one error happens on number 7, you can't recall 1 - 6.   Error handling here isn't in the scope of the issues I'm trying to tackle, but it is something we should maybe improve.

To overcome the limitations of the semantics of this call, I now have implemented a synchronous send method, which won't succeed unless all are sent, and you get back RecordMetadata.  This is what streams.publish will now use, which has the nice side effect of permitting us to return record metadata from kafka.

producer/src/main/kotlin/streams/procedures/StreamsProcedures.kt
moxious: Pushed updates.  @conker84 had implemented "fire and forget".  I changed the return types of several functions to return Futures.  The fire and forget behavior of this method has been retained, **but** I have implemented a new sendSync method according to @conker84 's suggestion.  I changed streams.publish to use this, so now it can return record metadata to the client.

"Fire and forget" is still possible with the API, and at a later point we might want to make this a config option.  Default behavior for streams.publish (in my opinion) should be to wait until it can confirm it did what you asked.  This matches the behavior of streams.consume which polls / blocks.

producer/src/main/kotlin/streams/kafka/KafkaEventRouter.kt
```diff
@@ -42,43 +43,44 @@ class KafkaEventRouter: StreamsEventRouter {
...
 
-    override fun sendEvents(topic: String, transactionEvents: List<out StreamsEvent>) {
-        try {
+    override fun sendEvents(topic: String, transactionEvents: List<out StreamsEvent>): List<out Future<RecordMetadata>> {
+        val sendFutures: MutableList<Future<RecordMetadata>> = mutableListOf()
+
+         try {
             producer.beginTransaction()
             transactionEvents.forEach {
+                // Use a mutable list and add each one, because we might send some and get an exception
```
moxious: TODO -- how should sendEvents behave if you fire & forget 6, but the third fails?

This change keeps present behavior intact, which is that it tries to do all 6.  A failure may stop some intermediate result, but it still does "best effort" on all of them.

producer/src/main/kotlin/streams/procedures/StreamsProcedures.kt
```diff
@@ -37,7 +62,9 @@ class StreamsProcedures {
                         .firstOrNull())
                 .withTopic(topic)
                 .build()
-        StreamsProcedures.eventRouter.sendEvents(topic, listOf(streamsEvent))
+        val result = StreamsProcedures.eventRouter.sendEventsSync(topic, listOf(streamsEvent))
```
moxious: Now using synchronous behavior by default which permits sending RecordMetadata.

producer/src/test/kotlin/streams/integrations/KafkaEventRouterIT.kt
```diff
@@ -166,6 +182,17 @@ class KafkaEventRouterIT {
         consumer.close()
     }
 
+    @Test
+    fun testCantPublishNull() {
```
moxious: @conker84  in a comment said that we should document this behavior, so now we throw an Exception and verify that is the case.

I wonder if we should allow publishing null though....I hear some people use this in kafka as a "tombstone record", but not sure if that applies here.  I'm new to this repo so I'm sticking with current behavior until we have a different requirement.

producer/src/main/kotlin/streams/procedures/StreamsProcedures.kt
conker84: I was wondering if instead of returning the `RecordMetadata`, it would be better to return a meta-structure that should be general as possible. This because we designed the Stream module in order to provide different "streaming connectors" (kinesis, rabbitMQ, ...)
@jexp @moxious wdyt?

moxious: If you like, a solution here is I could return record metadata as a map.  In this way I can include all the kafka-specific metadata, but the return type of the procedure can always be the same.  If it's a different streaming tech, you still get recordMetadata as a result, but the map has different keys.

conker84: Cool!

moxious: @jexp @conker84 I'm going down a deep rabbit hole here and might need some help.

So I created a wrapper class PublishedRecordMetadata and then implemented that in a specific case for kafka to make this generic.  However the Kafka publish functions all return Future<RecordMetadata>.   Doing research, these Futures are not pipe-lineable.  (And so doing the trivial bit of turning Future<RecordMetadata> into Future<KafkaPublishedRecordMetadata> is turning out to be stupid complex)

 In the Java world they're telling me that I need a CompletableFuture because Future is the old API.  In the Kotlin world they're telling me that I should use something else entirely.

So basically this simple suggestion has turned into something much more complicated for me with Kotlin/java async programming.

What is your opinion on how this should be done?  Forgetting for a moment the RecordMetadata class, what async implementation do you want this library to return e.g. return type here:

```
fun send(producerRecord: ProducerRecord<String, ByteArray>): Future<PublishedRecordMetadata>
```

I don't want to prep the PR again and find out that it needs to be totally different . :)

conker84: Maybe we can consider to keep it simple and change the _old_ `sendEvent` method in orderd to take a `callback` as third argument (so it can still be used with the fire-and-forget style); at this point the new `sendEventSync` should simply returns a `List<PublishedRecordMetadata>`

wdyt?

moxious: Not sure I understand how that would work.   Suppose I took a callback, I'm still going to get a `Future<RecordMetadata>` from the Kafka API.  If I write some code to `.get()` the value from that future, it's going to block in the async method -- or else if it doesn't block, then passing a callback is really the same as just calling `.get()` on the future yourself.

Isn't it also sort of the case that adding callbacks to APIs tends to make them messier and encourage "pyramid" style programming, I think Futures and promises were intended to get around this.   It does seem like a nasty combination of Java/Kotlin issues here though.

producer/src/main/kotlin/streams/procedures/StreamsProcedures.kt
```diff
@@ -15,15 +40,15 @@ class StreamsProcedures {
...
         checkEnabled()
         if (topic.isNullOrEmpty()) {
             log?.info("Topic empty, no message sent")
-            return
+            return Stream.empty();
         }
         if (payload == null) {
             log?.info("Payload empty, no message sent")
-            return
+            throw RuntimeException("Payload may not be null")
```
jexp: don't do that. imagine you have many messages then one null will result in a half-send batch

producer/src/test/kotlin/streams/mocks/MockStreamsEventRouter.kt
```diff
@@ -1,15 +1,44 @@
...
+            val future = CompletableFuture<RecordMetadata>()
+            future.complete(fakeRecordMetadata(topic))
+            result.add(future)
+        }
+
+        return result
+    }
+
+    override fun sendEventsSync(topic: String, transactionEvents: List<out StreamsEvent>): List<out RecordMetadata> {
+        var result = mutableListOf<RecordMetadata>()
```
jexp: transacations.map

producer/src/test/kotlin/streams/mocks/MockStreamsEventRouter.kt
```diff
@@ -1,15 +1,44 @@
...
+    fun fakeRecordMetadata(topic: String): RecordMetadata {
+        return RecordMetadata(
+                TopicPartition(topic, 0),
+                0, 1, 2, 3, 4, 5)
+    }
+
+    override fun sendEvents(topic: String, streamsTransactionEvents: List<out StreamsEvent>): List<out Future<RecordMetadata>> {
         events.addAll(streamsTransactionEvents as List<StreamsTransactionEvent>)
+
+        val result = mutableListOf<Future<RecordMetadata>>()
```
jexp: streamsTranscations.map

producer/src/test/kotlin/streams/mocks/MockStreamsEventRouter.kt
```diff
@@ -1,15 +1,44 @@
...
 import streams.events.StreamsEvent
 import streams.events.StreamsTransactionEvent
+import java.util.concurrent.CompletableFuture
+import java.util.concurrent.Future
 
 class MockStreamsEventRouter(logService: LogService? = null, config: Config? = null): StreamsEventRouter(logService, config) {
 
-    override fun sendEvents(topic: String, streamsTransactionEvents: List<out StreamsEvent>) {
+    fun fakeRecordMetadata(topic: String): RecordMetadata {
+        return RecordMetadata(
```
jexp: no need for return type + `return` just `= RecordMetadata(..`

producer/src/main/kotlin/streams/kafka/KafkaEventRouter.kt
```diff
@@ -42,43 +43,44 @@ class KafkaEventRouter: StreamsEventRouter {
...
         }
         val producerRecord = ProducerRecord(topic, partition, System.currentTimeMillis(), "${event.meta.txId + event.meta.txEventId}-${event.meta.txEventId}",
                 JSONUtils.writeValueAsBytes(event))
-        send(producerRecord)
+        return send(producerRecord)
     }
 
-    override fun sendEvents(topic: String, transactionEvents: List<out StreamsEvent>) {
-        try {
+    override fun sendEvents(topic: String, transactionEvents: List<out StreamsEvent>): List<out Future<RecordMetadata>> {
```
jexp: Instead using futures I would want to keep this `async` and then use `use await sendEvents()` in the sync method.
@conker84 will look into this.

producer/src/main/kotlin/streams/procedures/StreamsProcedures.kt
conker84: I think we should convert the `Future` in a `Deferred` by using (the Kotlin extension function [feature](https://kotlinlang.org/docs/reference/extensions.html)) in this way:
```kotlin
fun <T> Future<T>.asDeferred(dispatcher: CoroutineDispatcher = Dispatchers.Default): Deferred<T> {
    return GlobalScope.async(dispatcher) {
        get()
    }
}
```

An use it in that way:
```kotlin
private fun send(producerRecord: ProducerRecord<String, ByteArray>): Deferred<RecordMetadata> {
  return producer.send(producerRecord) { meta: RecordMetadata?, error: Exception? ->
    if (log.isDebugEnabled) {
      log.debug("Sent record in partition ${meta?.partition()} offset ${meta?.offset()} data ${meta?.topic()} key size ${meta?.serializedKeySize()}", error)
    }
  }.asDeferred()
}
```

conker84: I had a chat on the Kotlin Slack and basically, we have two options:
- change the implementation of `asDeferred` in this way:
```
fun <T> Future<T>.asDeferred(): Deferred<T> {
    return GlobalScope.async(Dispatchers.IO) {
        get()
    }
}
```
- implement a solution as described [here](https://stackoverflow.com/questions/50793362/how-to-use-suspendcoroutine-to-turn-java-7-future-into-kotlin-suspending-functio#answer-52901712)

moxious:
Too many issues, aged out.

PR #176, opened 2019-05-15
fixes 160: change the streams.sink.enabled to false

### LINKED ISSUE #160: change the streams.sink.enabled to false
The sink needs to be configured in combination with the topic/query mapping, so false should be the default value. Thanks to @moxious

Also evaluate to simply start the consumer in case almost one topic/query mapping option is provided.

conker84:
Fixes #160 

  - changed the `streams.sink.enabled` default value to `false`
  - updated the documentation
  - updated the docker-compose file


jexp:
Thanks a lot.

Issue #99, opened 2018-12-07
Provide a roundtrip-sink-config


jexp:
a default implementation of the sink that consumes source(producer) records

ingest data from CDC from Neo4j direction.

so for this one you would configure:

`streams.sink.topic.cdc.<TOPIC_NAME>=true`

or, whatever works best

`streams.sink.topic.cdc=topic-list`

PR #169, opened 2019-04-02
fixes #99: Provide a roundtrip-sink-config

conker84:
Fixes #99 

This PR is to provide two ways to ingest the data that comes from the CDC event.

## Proposed Changes (Mandatory)

A brief list of proposed changes in order to fix the issue:

 - Added `IngestionStrategy` in order to provide the ingestion by a strategy
   - Added the `MergeIngestionStrategy` where the data is merged by the Neo4j entity physical `id`.
   - Added the `SchemaIngestionStrategy` where the data is merged by the Neo4j constraints defined in the `schema` filed
 - created the class `StreamsSinkService` that routes the sink event

PR #178, opened 2019-05-19
 fixes #102: Manual commit behavior for handling errors and retrievals

### LINKED ISSUE #102: Manual commit behavior for handling errors and retrievals
To handle retrieval from a commit ID or from the beginning.
And to allow errors selectively commit or not of Kafka events.

That is something that we can do but we have to change the following auto-commit behaviour to a manual management like spark does:

https://github.com/apache/spark/blob/5264164a67df498b73facae207eda12ee133be7d/external/kafka-0-10-sql/src/main/scala/org/apache/spark/sql/kafka010/KafkaOffsetReader.scala#L132-L158

conker84:
Fixes #102 

Added Manual commit behaviour for handling errors and retrievals.
In case the property `kafka.enable.auto.commit` is set to false the consumer will be configured in order to manually commit the offsets.

  - Added new class `KafkaManualCommitEventConsumer` that manually commits the consumers
  - Added a new method in `StreamsEventConsumer`: `abstract fun read(topicConfig: Map<String, Any> = emptyMap(), action: (String, List<Any>) -> Unit)` that allow to pass a config map about reading a specific partition and starting offset and providing decision if commit or not the reading
  - added new configuraration parameters into the `streams.consume` procedure that allow more finegraded control over kafka connection

### DIFF
--- a/consumer/src/main/kotlin/streams/kafka/KafkaEventSink.kt
+++ b/consumer/src/main/kotlin/streams/kafka/KafkaEventSink.kt
@@ -1,34 +1,45 @@
 package streams.kafka
 
 import kotlinx.coroutines.*
-import org.apache.kafka.clients.consumer.KafkaConsumer
+import org.apache.kafka.clients.consumer.*
+import org.apache.kafka.common.TopicPartition
 import org.neo4j.kernel.configuration.Config
 import org.neo4j.logging.Log
 import streams.*
+import streams.kafka.KafkaTopicConfig.Companion.toTopicPartitionMap
 import streams.serialization.JSONUtils
 import streams.utils.StreamsUtils
+import java.util.concurrent.ConcurrentHashMap
+
 
 class KafkaEventSink(private val config: Config,
                      private val queryExecution: StreamsEventSinkQueryExecution,
                      private val streamsTopicService: StreamsTopicService,
                      private val log: Log): StreamsEventSink(config, queryExecution, streamsTopicService, log) {
 
-    private lateinit var eventConsumer: StreamsEventConsumer<*>
+    private lateinit var eventConsumer: StreamsEventConsumer
     private lateinit var job: Job
 
     private val streamsConfigMap = config.raw.filterKeys {
         it.startsWith("kafka.") || (it.startsWith("streams.") && !it.startsWith("streams.sink.topic.cypher."))
     }.toMap()
 
-    private val mappingKeys = mapOf("timeout" to "streams.sink.polling.interval",
-            "from" to "kafka.auto.offset.reset")
+    private val mappingKeys = mapOf(
+            "zookeeper" to "kafka.zookeeper.connect",
+            "broker" to "kafka.${ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG}",
+            "from" to "kafka.${ConsumerConfig.AUTO_OFFSET_RESET_CONFIG}",
+            "autoCommit" to "kafka.${ConsumerConfig.ENABLE_AUTO_COMMIT_CONFIG}",
+            "groupId" to "kafka.${ConsumerConfig.GROUP_ID_CONFIG}")
 
     override fun getEventConsumerFactory(): StreamsEventConsumerFactory {
         return object: StreamsEventConsumerFactory() {
-            override fun createStreamsEventConsumer(config: Map<String, String>, log: Log): StreamsEventConsumer<*> {
+            override fun createStreamsEventConsumer(config: Map<String, String>, log: Log): StreamsEventConsumer {
                 val kafkaConfig = KafkaSinkConfiguration.from(config)
-                val kafkaConsumer = KafkaConsumer<String, ByteArray>(kafkaConfig.asProperties())
-                return KafkaEventConsumer(kafkaConsumer, kafkaConfig.streamsSinkConfiguration, log)
+                return if (kafkaConfig.enableAutoCommit) {
+                    KafkaAutoCommitEventConsumer(kafkaConfig, log)
+                } else {
+                    KafkaManualCommitEventConsumer(kafkaConfig, log)
+                }
             }
         }
     }
@@ -71,12 +82,11 @@ class KafkaEventSink(private val config: Config,
         return GlobalScope.launch(Dispatchers.IO) {
             try {
                 while (isActive) {
-                    val data= eventConsumer.read()
-                    data?.forEach {
+                    eventConsumer.read { topic, data ->
                         if (log.isDebugEnabled) {
-                            log.debug("Reading data from topic ${it.key}, with data ${it.value}")
+                            log.debug("Reading data from topic $topic")
                         }
-                        queryExecution.writeForTopic(it.key, it.value)
+                        queryExecution.writeForTopic(topic, data)
                     }
                 }
                 eventConsumer.stop()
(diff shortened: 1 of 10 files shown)


mneedham:
@conker84 can we put the commit that fixes #99  into its own PR? I'm not that familiar with the code base yet so it'll be much easier for me to review them separately

mneedham:
@conker84 can we put the commit that fixes #99  into its own PR? I'm not that familiar with the code base yet so it'll be much easier for me to review them separately

I looked over the stuff the fixes #102 and that looks mostly fine to me. I wrote a couple of comments above.

mneedham:
The fixes for #99 looks good to me as well, but that one is bigger so best for @jexp to review it as well in case I've missed something

mneedham:
I see @jexp  has written most of the things I did about the #99 fixes on the other thread, so can ignore them here

conker84:
> Oh I see the one that fixes #99 is its own PR since April (#169) so is it on here by mistake or was that intentional?

I created this branch from the #99 so it should be easier rebase the code after that one will be merged

jexp:
Can you resolve the conflicts?

conker84:
Done!

sintezcs:
Hey guys, do you have any plans for merging this PR and releasing the new version of Kafka Plugin?

jexp:
In general:

We should split the `consumer/src/test/kotlin/integrations/KafkaEventSinkIT.kt` so it doesn't conflict so often with different features.

conker84:
I

> In general:
> 
> We should split the `consumer/src/test/kotlin/integrations/KafkaEventSinkIT.kt` so it doesn't conflict so often with different features.

I created the #184

jexp:
@sintezcs yes that's the plan :)

consumer/src/main/kotlin/streams/kafka/KafkaEventSink.kt
```diff
@@ -113,15 +148,172 @@ class KafkaEventConsumer(private val consumer: KafkaConsumer<String, ByteArray>,
...
+            } catch (e: Exception) {
+                // TODO add dead letter queue
+            }
+        }
+    }
+
+    private fun readFromPartition(config: KafkaTopicConfig, action: (String, List<Any>) -> Unit) {
+        setSeek(config.topicPartitionsMap)
+        val records = consumer.poll(0)
+        val consumerRecordsMap = toConsumerRecordsMap(config.topicPartitionsMap, records)
```
jexp: why is the `records` handling here different from the readSimple method after `consumer.poll()`
one would guess it's the same kind of records.
I would kinda expect here only the setSeek as the only difference between the two methods?

consumer/src/main/kotlin/streams/kafka/KafkaEventSink.kt
```diff
@@ -113,15 +148,172 @@ class KafkaEventConsumer(private val consumer: KafkaConsumer<String, ByteArray>,
...
+        if (consumerRecordsMap.isNotEmpty()) {
```
jexp: aren't the empty ones already filtered out?

consumer/src/main/kotlin/streams/kafka/KafkaEventSink.kt
```diff
@@ -113,15 +148,172 @@ class KafkaEventConsumer(private val consumer: KafkaConsumer<String, ByteArray>,
...
+                }
+                .filterValues { it.isNotEmpty() }
+
+    fun setSeek(topicPartitionsMap: Map<TopicPartition, Long>) {
+        if (isSeekSet) {
+            return
+        }
+        isSeekSet = true
+        consumer.poll(0) // dummy call see: https://stackoverflow.com/questions/41008610/kafkaconsumer-0-10-java-api-error-message-no-current-assignment-for-partition
+        topicPartitionsMap.forEach {
```
jexp: some naming would be good here, e.g. it.key == topic and it.value -> seek

consumer/src/main/kotlin/streams/kafka/KafkaEventSink.kt
```diff
@@ -113,15 +148,172 @@ class KafkaEventConsumer(private val consumer: KafkaConsumer<String, ByteArray>,
...
+        } else {
+            readFromPartition(kafkaTopicConfig, action)
+        }
+    }
+
+    fun toConsumerRecordsMap(topicPartitionsMap: Map<TopicPartition, Long>,
+                             records: ConsumerRecords<String, ByteArray>)
+            : Map<TopicPartition, List<ConsumerRecord<String, ByteArray>>> = topicPartitionsMap
+                .mapValues {
+                    records.records(it.key)
```
jexp: naming would be good here, I think you can deconstruct pairs as `.mapValues { (topic, partition) -> }`

jexp: https://play.kotlinlang.org/#eyJ2ZXJzaW9uIjoiMS4zLVJDIiwicGxhdGZvcm0iOiJqYXZhIiwiYXJncyI6IiIsIm5vbmVNYXJrZXJzIjp0cnVlLCJ0aGVtZSI6ImlkZWEiLCJmb2xkZWRCdXR0b24iOnRydWUsInJlYWRPbmx5IjpmYWxzZSwiY29kZSI6ImZ1biBtYWluKCkgeyBcbnZhbCBhID0gbGlzdE9mKFwiYVwiIHRvIDIsXCJiXCIgdG8gMykudG9NYXAoKVxuXG52YWwgYiA9IGEubWFwVmFsdWVzeyAoa2V5LCB2YWx1ZSkgLT4gXG4gICAgdmFsdWUgKzFcbn1cbnByaW50bG4oYilcblxufSJ9

consumer/src/main/kotlin/streams/kafka/KafkaEventSink.kt
```diff
@@ -113,15 +148,172 @@ class KafkaEventConsumer(private val consumer: KafkaConsumer<String, ByteArray>,
...
+                else -> consumer.seek(it.key, it.value)
+            }
+        }
+    }
+}
+
+class KafkaManualCommitEventConsumer(private val config: KafkaSinkConfiguration,
+                                     private val log: Log): KafkaAutoCommitEventConsumer(config, log) {
+
+    private val topicPartitionOffsetMap = ConcurrentHashMap<TopicPartition, OffsetAndMetadata>()
```
jexp: will this grow indefinitely?

consumer/src/main/kotlin/streams/kafka/KafkaEventSink.kt
```diff
@@ -113,15 +148,172 @@ class KafkaEventConsumer(private val consumer: KafkaConsumer<String, ByteArray>,
...
+        val topicMap = mutableMapOf<TopicPartition, OffsetAndMetadata>()
+        val records = consumer.poll(0)
+        if (!records.isEmpty) {
+            this.topics.forEach { topic ->
+                val topicRecords = records.records(topic)
+                val lastRecord = topicRecords.last()
+                val offsetAndMetadata = OffsetAndMetadata(lastRecord.offset(), "")
+                val topicPartition = TopicPartition(lastRecord.topic(), lastRecord.partition())
+                topicMap[topicPartition] = offsetAndMetadata
+                topicPartitionOffsetMap[topicPartition] = offsetAndMetadata
```
jexp: I would make this clearer that you're writing to a instance variable, e.g. by using a named method.

consumer/src/main/kotlin/streams/kafka/KafkaEventSink.kt
```diff
@@ -113,15 +148,172 @@ class KafkaEventConsumer(private val consumer: KafkaConsumer<String, ByteArray>,
...
+
+    override fun start() {
+        if (topics.isEmpty()) {
+            log.info("No topics specified Kafka Consumer will not started")
+            return
+        }
+        this.consumer.subscribe(topics, StreamsConsumerRebalanceListener(topicPartitionOffsetMap, consumer, config.autoOffsetReset, log))
+    }
+
+    private fun readSimple(action: (String, List<Any>) -> Unit): Map<TopicPartition, OffsetAndMetadata> {
```
jexp: these two methods have almost the same code, is there any way that we can unify them?

consumer/src/main/kotlin/streams/kafka/KafkaEventSink.kt
```diff
@@ -113,15 +148,172 @@ class KafkaEventConsumer(private val consumer: KafkaConsumer<String, ByteArray>,
...
+    }
+
+    private fun commitData(commit: Boolean, topicMap: Map<TopicPartition, OffsetAndMetadata>) {
+        if (commit && topicMap.isNotEmpty()) {
+            consumer.commitSync(topicMap)
+        }
+    }
+
+    override fun read(topicConfig: Map<String, Any>, action: (String, List<Any>) -> Unit) {
+        val kafkaTopicConfig = KafkaTopicConfig.fromMap(topicConfig)
```
jexp: do we have to do this for every read? wouldn't it make more sense to only compute it once? I guess the config doesn't change so much?

Only if we want to make the config reloadable etc.

consumer/src/main/kotlin/streams/procedures/StreamsSinkProcedures.kt
```diff
@@ -27,25 +29,40 @@ class StreamsSinkProcedures {
...
+    }
+
+    private fun readData(topic: String, procedureConfig: Map<String, Any>, consumerConfig: Map<String, String>): List<Any> {
+        val cfg = procedureConfig.mapValues { if (it.key != "partitions") it.value else mapOf(topic to it.value) }
+        val timeout = cfg.getOrDefault("timeout", 1000).toString().toLong()
+        val consumer = createConsumer(consumerConfig, topic)
         consumer.start()
-        val data = try {
-            consumer.read()
+        val data = mutableListOf<Any>()
```
jexp: I guess we'll have to rewrite that in the future into something async.
is consumer.read blocking?
if not this loop can burn a lot of cpu

PR #187, opened 2019-06-03
fixes #185: Consumer not working in Cluster environment

### LINKED ISSUE #185: Consumer not working in Cluster environment
Consumer not working in Cluster environment because of they are all in FOLLOWER state and the `Neo4jUtils.isWriteableInstance` blocks the write.
We have to wait that the RAFT election ends

conker84:
Fixes #185 

  - Wait for the leader election in case of a cluster environment

### DIFF
--- a/common/src/main/kotlin/streams/utils/Neo4jUtils.kt
+++ b/common/src/main/kotlin/streams/utils/Neo4jUtils.kt
@@ -1,11 +1,19 @@
 package streams.utils
 
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.GlobalScope
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.launch
 import org.neo4j.graphdb.QueryExecutionException
 import org.neo4j.kernel.impl.logging.LogService
 import org.neo4j.kernel.internal.GraphDatabaseAPI
+import org.neo4j.logging.Log
+import org.neo4j.logging.NullLog
 import java.lang.reflect.InvocationTargetException
+import kotlin.streams.toList
 
 object Neo4jUtils {
+    @JvmStatic val LEADER = "LEADER"
     fun isWriteableInstance(db: GraphDatabaseAPI): Boolean {
         try {
             val isSlave = StreamsUtils.ignoreExceptions(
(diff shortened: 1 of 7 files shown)


common/src/main/kotlin/streams/utils/Neo4jUtils.kt
```diff
@@ -32,4 +40,61 @@ object Neo4jUtils {
...
+            action()
+        }
+    }
+
+    fun executeInLeader(db: GraphDatabaseAPI, log: Log, timeout: Long = 50000, action: () -> Unit) {
+        if (isEnterpriseEdition(db)) {
+            // wait for the election
+            GlobalScope.launch(Dispatchers.IO) {
+                val start = System.currentTimeMillis()
+                var end = System.currentTimeMillis()
```
jexp: we don't have to store `end`

consumer/src/main/kotlin/streams/StreamsEventSinkExtensionFactory.kt
```diff
@@ -51,27 +51,21 @@ class StreamsEventSinkExtensionFactory : KernelExtensionFactory<StreamsEventSink
...
+                                        log,
+                                        db)
+                        // start the Sink
+                        Neo4jUtils.executeInLeader(db, log) { initSinkModule() }
 
                         // Register required services for the Procedures
                         StreamsSinkProcedures.registerStreamsSinkConfiguration(streamsSinkConfiguration)
                         StreamsSinkProcedures.registerStreamsEventConsumerFactory(eventSink.getEventConsumerFactory())
                         StreamsSinkProcedures.registerStreamsEventSinkConfigMapper(eventSink.getEventSinkConfigMapper())
-                        streamsLog.info("Streams Sink module initialised")
```
jexp: why did you remove this?

conker84: I moved it into the `initSinkModule` method

common/src/main/kotlin/streams/utils/Neo4jUtils.kt
```diff
@@ -32,4 +40,61 @@ object Neo4jUtils {
         return db.dependencyResolver
                 .resolveDependency(LogService::class.java)
     }
+
+    fun isEnterpriseEdition(db: GraphDatabaseAPI): Boolean {
+        try {
+            db.execute("CALL dbms.cluster.role()").columnAs<String>("role").next()
```
mneedham: For checking for version, we can use this procedure:

```
CALL dbms.components() YIELD edition
RETURN edition = "enterprise" AS isEnterprise
```

conker84: Thanks @mneedham I changed the method with your suggestion!

PR #190, opened 2019-06-04
fixes #186: Kafka even sink with manual commit fails with multiple topic subscriptions

### LINKED ISSUE #186: Kafka even sink with manual commit fails with multiple topic subscriptions
I've built the plugin from source to try the latest functionality from https://github.com/neo4j-contrib/neo4j-streams/pull/178.

Neo4J streams plugin fails to process the messages from multiple topics if `kafka.enable.auto.commit` is disabled. **If you leave only one topic in the configuration - everything works fine.**

## Expected Behavior
Neo4J can receive events from both `maxwell_project` and `maxwell_artist` topics and process them with the corresponding CYPHER queries.

## Actual Behavior
Neo4J raises an error when an event comes to any of the topics:
```
2019-06-03 10:34:57.406+0000 ERROR Collection is empty. Collection is empty.
java.util.NoSuchElementException: Collection is empty.
	at kotlin.collections.CollectionsKt___CollectionsKt.last(_Collections.kt:348)
	at streams.kafka.KafkaManualCommitEventConsumer.readSimple(KafkaEventSink.kt:229)
	at streams.kafka.KafkaManualCommitEventConsumer.read(KafkaEventSink.kt:276)
	at streams.StreamsEventConsumer.read$default(StreamsEventSink.kt:29)
	at streams.kafka.KafkaEventSink$createJob$1.invokeSuspend(KafkaEventSink.kt:85)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:32)
	at kotlinx.coroutines.DispatchedTask$DefaultImpls.run(Dispatched.kt:235)
	at kotlinx.coroutines.DispatchedContinuation.run(Dispatched.kt:81)
	at kotlinx.coroutines.scheduling.Task.run(Tasks.kt:94)
	at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:586)
	at kotlinx.coroutines.scheduling.CoroutineScheduler.access$runSafely(CoroutineScheduler.kt:60)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:732)
```

### Steps (Mandatory)

  1. Disable the auto-commit 
`kafka.enable.auto.commit=false`
  2. Subscribe to mutliple (more than 1) topics:
```
kafka.enable.auto.commit=false
streams.sink.enabled=true

streams.sink.topic.cypher.maxwell_artist=\
WITH event.data AS data \
    MERGE (ar:Artist {id: toInteger(data.artist_id)}) \
      ON CREATE SET ar.name = data.artist_name \
      ON MATCH SET ar.name = data.artist_name

streams.sink.topic.cypher.maxwell_project=\
WITH event.data AS data \
    MATCH (ar:Artist {id: toInteger(data.artist_id)}) \
    MERGE (pr:Project {id: toInteger(data.project_id)}) \
      ON CREATE SET pr.name = data.project_name \
      ON MATCH SET pr.name = data.project_name \
    MERGE (ar)-[:HAS_PROJECT]->(pr)
```

## Specifications

Currently used versions

### Versions

  - OS: MacOS  10.14.5
  - Neo4j: 3.4.1
  - Neo4j-Streams: 3.4.1 (built from sources)

conker84:
Fixes #186

  - Fix the bugs

### DIFF
--- a/consumer/src/main/kotlin/streams/kafka/KafkaEventSink.kt
+++ b/consumer/src/main/kotlin/streams/kafka/KafkaEventSink.kt
@@ -79,7 +79,7 @@ class KafkaEventSink(private val config: Config,
 
     private fun createJob(): Job {
         log.info("Creating Sink daemon Job")
-        return GlobalScope.launch(Dispatchers.IO) {
+        return GlobalScope.launch(Dispatchers.IO) { // TODO improve exception management
             try {
                 while (isActive) {
                     eventConsumer.read { topic, data ->
@@ -153,7 +153,11 @@ open class KafkaAutoCommitEventConsumer(private val config: KafkaSinkConfigurati
         if (!records.isEmpty) {
             try {
                 this.topics.forEach { topic ->
-                    action(topic, records.records(topic).map { JSONUtils.readValue<Any>(it.value()) })
+                    val topicRecords = records.records(topic)
+                    if (!topicRecords.iterator().hasNext()) {
+                        return@forEach
+                    }
+                    action(topic, topicRecords.map { JSONUtils.readValue<Any>(it.value()) })
                 }
             } catch (e: Exception) {
                 // TODO add dead letter queue
@@ -186,10 +190,10 @@ open class KafkaAutoCommitEventConsumer(private val config: KafkaSinkConfigurati
     fun toConsumerRecordsMap(topicPartitionsMap: Map<TopicPartition, Long>,
                              records: ConsumerRecords<String, ByteArray>)
             : Map<TopicPartition, List<ConsumerRecord<String, ByteArray>>> = topicPartitionsMap
-                .mapValues {
-                    records.records(it.key)
-                }
-                .filterValues { it.isNotEmpty() }
+            .mapValues {
+                records.records(it.key)
+            }
+            .filterValues { it.isNotEmpty() }
 
     fun setSeek(topicPartitionsMap: Map<TopicPartition, Long>) {
         if (isSeekSet) {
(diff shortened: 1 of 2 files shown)


consumer/src/main/kotlin/streams/kafka/KafkaEventSink.kt
```diff
@@ -153,7 +153,11 @@ open class KafkaAutoCommitEventConsumer(private val config: KafkaSinkConfigurati
         if (!records.isEmpty) {
             try {
                 this.topics.forEach { topic ->
-                    action(topic, records.records(topic).map { JSONUtils.readValue<Any>(it.value()) })
+                    val topicRecords = records.records(topic)
+                    if (!topicRecords.iterator().hasNext()) {
```
jexp: can't you just do the . action in the if, and not have the lambda return.
removes one negation and makes the code simpler.

jexp: also it's odd as it should just have called the action with an empty list ??

consumer/src/main/kotlin/streams/kafka/KafkaEventSink.kt
```diff
@@ -226,6 +230,9 @@ class KafkaManualCommitEventConsumer(private val config: KafkaSinkConfiguration,
         if (!records.isEmpty) {
             this.topics.forEach { topic ->
                 val topicRecords = records.records(topic)
+                if (!topicRecords.iterator().hasNext()) {
```
jexp: perhaps rather doe topics.map.filter.foreach ? 
if conditions with local returns within closures are imho a code-smell.

jexp:
My comments are not critical so we can merge this and fix those places later in a clean up.
WDYT @mneedham

consumer/src/main/kotlin/streams/kafka/KafkaEventSink.kt
conker84: In this case the check is a plus, we can omitted it, it's just because I don't want pass empty lists.
Wdyt if, as you suggested in the next comment, I do `topics.map.filter.foreach` also here?

jexp: yes, makes sense. easier to reason about too.

PR #206, opened 2019-06-24
fixes #208: Kafka Connect sink Topics Validation must not take count of topic order definition

conker84:
Fixes #208

Kafka Connect sink Topics Validation must not take count of topic order definition

  - lists sorted

### DIFF
--- a/kafka-connect-neo4j/src/test/kotlin/streams/kafka/connect/sink/Neo4jSinkConnectorConfigTest.kt
+++ b/kafka-connect-neo4j/src/test/kotlin/streams/kafka/connect/sink/Neo4jSinkConnectorConfigTest.kt
@@ -11,17 +11,17 @@ import kotlin.test.assertNull
 
 class Neo4jSinkConnectorConfigTest {
 
-   @Test(expected = ConfigException::class)
-   fun `should throw a ConfigException because of mismatch`() {
-       try {
-           val originals = mapOf(SinkConnector.TOPICS_CONFIG to "foo, bar",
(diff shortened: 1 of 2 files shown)


kafka-connect-neo4j/src/test/kotlin/streams/kafka/connect/sink/Neo4jSinkConnectorConfigTest.kt
```diff
@@ -11,17 +11,17 @@ import kotlin.test.assertNull
...
-       }
-   }
+    @Test(expected = ConfigException::class)
+    fun `should throw a ConfigException because of mismatch`() {
+        try {
+            val originals = mapOf(SinkConnector.TOPICS_CONFIG to "foo, bar",
+                    "${Neo4jSinkConnectorConfig.TOPIC_CYPHER_PREFIX}foo" to "CREATE (p:Person{name: event.firstName})")
+            Neo4jSinkConnectorConfig(originals)
+        } catch (e: ConfigException) {
+            assertEquals("There is a mismatch between topics defined into the property `topics` and configured topics ([foo])", e.message)
```
mneedham: Can we add to the error message so that it indicates which topics were defined for the property topics?

So for this example it might say:

`There is a mismatch between topics defined into the property `topics` ([foo,bar]) and configured topics ([foo])`

conker84: Sure!

PR #210, opened 2019-07-12
fixes #184: Split KafkaEventSinkIT

### LINKED ISSUE #184: Split KafkaEventSinkIT
We should split the consumer/src/test/kotlin/integrations/KafkaEventSinkIT.kt so it doesn't conflict so often with different features.

conker84:
Fixes #184

Splitted KafkaEventSinkIT in serveral classes:
 - KafkaEventSinkCDC
 - KafkaEventSinkCommit
 - KafkaEventSinkDLQ
 - KafkaEventSinkPattern
 - KafkaEventSinkSimple

Created a test suite that spins up the Kafka Stack just once in order to reduce the Container spin-up overhead.

  - Splitted KafkaEventSinkIT
  - Created a test suite that spins up the Kafka Stack just once in order to reduce
  - Splitted KafkaEventSink into different files


consumer/src/test/kotlin/integrations/kafka/KafkaEventSinkSimple.kt
```diff
@@ -0,0 +1,93 @@
...
+import org.junit.Test
+import org.neo4j.function.ThrowingSupplier
+import org.neo4j.graphdb.Node
+import org.neo4j.kernel.internal.GraphDatabaseAPI
+import org.neo4j.test.assertion.Assert
+import streams.serialization.JSONUtils
+import java.util.*
+import java.util.concurrent.TimeUnit
+
+class KafkaEventSinkSimple: KafkaEventSinkBase() {
```
mneedham: Should this one have `IT` as the suffix of the class?

consumer/src/test/kotlin/integrations/kafka/KafkaEventSinkPattern.kt
```diff
@@ -0,0 +1,90 @@
...
+import org.junit.Test
+import org.neo4j.function.ThrowingSupplier
+import org.neo4j.kernel.internal.GraphDatabaseAPI
+import org.neo4j.test.assertion.Assert
+import streams.serialization.JSONUtils
+import java.util.*
+import java.util.concurrent.TimeUnit
+import kotlin.test.assertEquals
+
+class KafkaEventSinkPattern : KafkaEventSinkBase() {
```
mneedham: This is a test as well right? `IT` suffix would make it clearer I think

consumer/src/test/kotlin/integrations/kafka/KafkaEventSinkSimple.kt
conker84: I removed the `IT` from all the tests invoked by `KafkaEventSinkSuiteIT` because otherwise they trigger this configuration:

https://github.com/neo4j-contrib/neo4j-streams/blob/b597f09918a4de345286de558d38aaee6b5f693a/pom.xml#L250-L258

And they would be executed twice.
Should we define a naming convention for all the tests involved in a suite? Something like `<NAME>Suite`, or whatever else.

consumer/src/test/kotlin/integrations/kafka/KafkaEventSinkPattern.kt
conker84: As well as above

PR #216, opened 2019-07-24
fixes #211: Add AVRO Support to the Streams Sink Plugin

### LINKED ISSUE #211: Add AVRO Support to the Streams Sink Plugin
The Kafka Connect Plugin natively supports the AVRO file format.
We need to add the same feature to the Streams Plugin

conker84:
Fixes #211

  - Added AVRO+Schema Registry support to sink and `streams.consume` procedure
  - updated documentation


common/src/main/kotlin/streams/extensions/CommonExtensions.kt
```diff
@@ -34,9 +39,45 @@ fun Map<String, Any?>.flatten(map: Map<String, Any?> = this, prefix: String = ""
...
-fun <K, V> ConsumerRecord<K, V>.offsetAndMetadata(metadata: String = "") = OffsetAndMetadata(this.offset() + 1, metadata)
+fun ConsumerRecord<*, *>.topicPartition() = TopicPartition(this.topic(), this.partition())
+fun ConsumerRecord<*, *>.offsetAndMetadata(metadata: String = "") = OffsetAndMetadata(this.offset() + 1, metadata)
 
-fun ConsumerRecord<ByteArray, ByteArray>.toStreamsSinkEntity(): StreamsSinkEntity = StreamsSinkEntity(
-        if (this.key() == null) null else JSONUtils.readValue<Any>(this.key(), true),
-        if (this.value() == null) null else JSONUtils.readValue<Any>(this.value()))
\ No newline at end of file
+private fun convertAvroData(rawValue: Any?): Any? = when (rawValue) {
+    is IndexedRecord -> rawValue.toMap()
```
jexp: is this handling nested records/arrays of records in the map?

common/src/main/kotlin/streams/extensions/CommonExtensions.kt
```diff
@@ -34,9 +39,45 @@ fun Map<String, Any?>.flatten(map: Map<String, Any?> = this, prefix: String = ""
...
+    is Collection<*> -> rawValue.toList().map { convertAvroData(it) }
```
jexp: is there something else for Arrays that we need to cover?

common/src/main/kotlin/streams/extensions/CommonExtensions.kt
jexp: see below for `is Map`

common/src/main/kotlin/streams/extensions/CommonExtensions.kt
```diff
@@ -34,9 +39,45 @@ fun Map<String, Any?>.flatten(map: Map<String, Any?> = this, prefix: String = ""
...
+        .map { it.name() to convertAvroData(this[it.pos()]) }
+        .toMap()
+
+fun Schema.toMap() = JSONUtils.asMap(this.toString())
+
+private fun convertData(data: Any?): Any? {
+    return data?.let {
+        when (data) {
+            is ByteArray -> JSONUtils.readValue<Any>(data)
+            is GenericRecord -> data.toMap()
```
jexp: what happens here with nested values in the map (arrays, collections of records or singluar records etc)?

common/src/main/kotlin/streams/extensions/CommonExtensions.kt
```diff
@@ -34,9 +39,45 @@ fun Map<String, Any?>.flatten(map: Map<String, Any?> = this, prefix: String = ""
...
+            else -> throw RuntimeException("Unsupported type ${data::class.java.name}")
+        }
+    }
+}
+fun ConsumerRecord<*, *>.toStreamsSinkEntity(): StreamsSinkEntity {
+    val rawKey = this.key()
+    val key = try {
+        convertData(rawKey)
+    } catch(e: Exception) {
```
jexp: is Exception here too generic, what about that runtime exception from above about the unsupported type? should we turn that value really into a string ? or fail?

conker84: what about something like this?

```Kotlin
private fun convertData(data: Any?, stringWhenFailure: Boolean = false): Any? {
    return when (data) {
        is ByteArray -> JSONUtils.readValue<Any>(data)
        is GenericRecord -> data.toMap()
        null -> null
        else -> if (stringWhenFailure) data.toString() else throw RuntimeException("Unsupported type ${data::class.java.name}")
    }
}
fun ConsumerRecord<*, *>.toStreamsSinkEntity(): StreamsSinkEntity {
    val key = convertData(this.key(), true)
    val value = convertData(this.value())
    return StreamsSinkEntity(key, value)
}
```

jexp:
Needs to be rebased again, sorry.

common/src/main/kotlin/streams/extensions/CommonExtensions.kt
jexp: yep

jexp:
and the nested conversion has to be handled correctly (if needed).

PR #224, opened 2019-08-09
fixes #188: Support Specialized Sink Classes: support CUD format

### LINKED ISSUE #188: Support Specialized Sink Classes: support CUD format

Specialized Sink Classes.

Right now, I usually do this:

```NEO4J_streams_sink_topic_cypher_bank: "MERGE (b:Bank { id: event.id }) SET b += event"```

What I'd really like to do is this:

```NEO4J_streams_sink_topic_cypher_bank: org.foo.whatever.GenericSink```

This would do THREE things for me:
1. Not repeat the same formulaic cypher over and over for many different label/topic types
2. Allow me to implement custom sink types
3. Allow more complex config later that is expressed as a map.  Imagine for example:

NEO4J_streams_sink_mapping:  topic1:GenericSink,topic2:CustomSink,topic3:GenericSink

GenericSink would implement some interface.   If we did this, then we could think of the actual standard Cypher route as actually being an invocation of a "CypherSink", or maybe "CypherMessageConsumer" or something like that.

## Considered alternatives

The main alternative is developing special syntaxes for the config file.  These I think we should be careful to avoid because of the "syntax within a syntax" problem of the configuration file.  Plus inventing new syntax will lead users to ask for more features, until you're soon back at the complexity of using cypher, which I'd like to avoid in a config file.

Make it easier to start out and configure the module, less Cypher error prone, and more extensible.

### PR DESCRIPTION
**Pull request by conker84**
Fixes #188

  - Added the support to the CUD Format
  - Updated Documentation

### CODE REVIEW

File: common/src/main/kotlin/streams/service/sink/strategy/CUDIngestionStrategy.kt [hunk 1]
```diff
@@ -0,0 +1,263 @@
...
+        return when (ids.contains("_id")) {
+            true -> "MATCH ($identifier:${getLabelsAsString(labels)}) WHERE id($identifier) = event.$fullField$ID_KEY._id"
+            else -> "$keyword ($identifier:${getLabelsAsString(labels)} {${getNodeKeysAsString(keys = ids, prefix = "$fullField$ID_KEY")}})"
+        }
+    }
+
+    private fun buildNodeCreateStatement(labels: List<String>): String = """
+            |${StreamsUtils.UNWIND}
+            |CREATE (n:${getLabelsAsString(labels)})
+            |SET n = event.properties
```
moxious: minor point - but in these generated cypher statements you're generally matching nodes to n and relationships to r, which makes sense, but in all of these strategies you're operating on some graph element or "atom".  It might be good to match them all to the same variable.  

Just as kafka guarantees that what goes in will always be called "event", a CUD strategy might guarantee that what comes out will always be called "element" (for example).  Later on this would help if we choose to chain this strategy to something else.

common/src/main/kotlin/streams/service/sink/strategy/CUDIngestionStrategy.kt
```diff
@@ -0,0 +1,263 @@
...
+            |${buildNodeLookupByIds(ids = from.ids, labels = from.labels, identifier = "from", field = "from")}
+            |${buildNodeLookupByIds(ids = to.ids, labels = to.labels, identifier = "to", field = "to")}
+            |MATCH (from)-[r:${rel_type.quote()}]->(to)
+            |SET r += event.properties
+        """.trimMargin()
+
+    private fun buildDeleteStatement(labels: List<String>, ids: Set<String>, detach: Boolean): String = """
+            |${StreamsUtils.UNWIND}
+            |${buildNodeLookupByIds(ids = ids, labels = labels)}
+            |${if (detach) "DETACH " else ""}DELETE n
```
moxious: Hold on don't you always have to detach?  This is maybe a behavior quesiton.

If you don't always DETACH then if a user tries to delete a connected node it will fail and then you have a special case to handle.

This is probably a fine point to document, but I think we should always DETACH DELETE the node to guarantee that what the user asked for (delete the node) actually happens.   Then put in the documentation that "If you do this, it implies you're deleting the incident relationships too" because it has to be that way.

common/src/main/kotlin/streams/service/sink/strategy/CUDIngestionStrategy.kt
```diff
@@ -0,0 +1,263 @@
...
+    companion object {
+        @JvmStatic val ID_KEY = "ids"
+    }
+
+    data class NodeRelMetadata(val labels: List<String>, val ids: Set<String>)
+
+    private fun buildNodeLookupByIds(keyword: String = "MATCH", ids: Set<String>, labels: List<String>, identifier: String = "n", field: String = ""): String {
+        val fullField = if (field.isNotBlank()) "$field." else field
+        return when (ids.contains("_id")) {
+            true -> "MATCH ($identifier:${getLabelsAsString(labels)}) WHERE id($identifier) = event.$fullField$ID_KEY._id"
```
moxious: you have to escape all identifiers and basically ALL inputs to this function with backticks.   If I pass an identifier with a space in it, or a special character, this will break.

Double check code in getLabelsAsString for the same.

common/src/main/kotlin/streams/service/sink/strategy/CUDIngestionStrategy.kt
```diff
@@ -0,0 +1,263 @@
...
+                .groupBy { Triple(it.labels, it.ids.keys, it.detach) }
+                .map {
+                    val (labels, keys, detach) = it.key
+                    QueryEvents(buildDeleteStatement(labels, keys, detach), it.value.map { it.toMap() })
+                }
+    }
+
+    override fun mergeRelationshipEvents(events: Collection<StreamsSinkEntity>): List<QueryEvents> {
+        val data = events
+                .mapNotNull {
```
moxious: you have a couple of functions like this which are mapping through all of the events of a certain CUD type.

Same comment on all of them -- you're doing some data validation like the event has to be not null, the operation has to make sense, all of that's good.  But this code is dropping on the floor any kafka message that violates its assumptions.

This is a spot where it'd be good to have a debug level statement saying "Hey, I got some messages but I'm dropping them on the floor because they don't apply or don't make sense to this scenario".   When there's not an option to get this debugging output, the behavior of the plugin may appear buggy or mysterious.

common/src/test/kotlin/streams/service/sink/strategy/CUDIngestionStrategyTest.kt
```diff
@@ -0,0 +1,572 @@
...
+        assertEquals(1, updateNodeFooBarLabel.events.size)
+        assertNodeEventsContainsKey(updateNodeFooBarLabel, key)
+    }
+
+    @Test
+    fun `should delete nodes`() {
+        // given
+        val detachMarkers = listOf(1, 3, 8, 10)
+        val list = (1..10).map {
+            val labels = if (it % 2 == 0) listOf("Foo", "Bar") else listOf("Foo", "Bar", "Label")
```
moxious: In your test cases, please make sure to use labels with garbage data and also properties with the same, because this will happen in the real world.   For example:

labels = ["WellBehaved", "C̸r̵a̵z̵y̵ ̶.̵ ̶ ̴ ̸ ̶ ̶ ̵ ̴L̴a̵b̸e̶l̴"]
properties = ["simple", "not..... SO SIMPLE!"]

common/src/test/kotlin/streams/service/sink/strategy/CUDIngestionStrategyTest.kt
```diff
@@ -0,0 +1,572 @@
...
+        val relationshipDeleteEvents = cudQueryStrategy.deleteRelationshipEvents(list)
+
+        // then
+        assertEquals(emptyList(), nodeEvents)
+        assertEquals(emptyList(), relationshipEvents)
+        assertEquals(emptyList(), relationshipDeleteEvents)
+
+        assertEquals(4, nodeDeleteEvents.size)
+        assertEquals(10, nodeDeleteEvents.map { it.events.size }.sum())
+        val deleteNodeFooBar = findEventByQuery("""
```
moxious: Recommend testing the deletion by a compound key (>= 2 properties).

common/src/test/kotlin/streams/service/sink/strategy/CUDIngestionStrategyTest.kt
```diff
@@ -0,0 +1,572 @@
...
+                |${StreamsUtils.UNWIND}
+                |MATCH (n:Foo:Bar:Label {key: event.${CUDIngestionStrategy.ID_KEY}.key})
+                |DETACH DELETE n
+            """.trimMargin(), nodeDeleteEvents)
+        assertEquals(2, deleteNodeFooBarLabelDetach.events.size)
+        assertNodeEventsContainsKey(deleteNodeFooBarLabelDetach, "key")
+    }
+
+    @Test
+    fun `should delete nodes with internal id reference`() {
```
moxious: Functionality / behavior question.  Not sure I saw this spot in the code, but what happens if the message tries to delete something by an internal ID which obviously isn't an internal ID?   Like id="banana".

This goes back to the earlier point about debug messaging when the plugin has to drop certain records because they're malformed.

common/src/test/kotlin/streams/service/sink/strategy/CUDIngestionStrategyTest.kt
```diff
@@ -0,0 +1,572 @@
...
+                |${StreamsUtils.UNWIND}
+                |MATCH (n:Foo:Bar) WHERE id(n) = event.${CUDIngestionStrategy.ID_KEY}._id
+                |DELETE n
+            """.trimMargin(), nodeDeleteEvents)
+        assertEquals(3, deleteNodeFooBar.events.size)
+        val key = "_id"
+        assertNodeEventsContainsKey(deleteNodeFooBar, key)
+        val deleteNodeFooBarDetach = findEventByQuery("""
+                |${StreamsUtils.UNWIND}
+                |MATCH (n:Foo:Bar) WHERE id(n) = event.${CUDIngestionStrategy.ID_KEY}._id
```
moxious: See earlier comment.  If we generate cypher like this and _id = "banana", then the plugin will successfully process the message, nothing will happen, and the user will think the plugin is very mysterious.  :)

Some upfront message validation may be necessary

kafka-connect-neo4j/docker/readme.adoc
```diff
@@ -497,6 +499,122 @@ By default no properties will be attached to the edge nodes but you can specify
...
+  },
+  "to": {
+    "ids": {"otherKey":1},
+    "labels": ["FooBar"]
+  },
+  "type":"relationship"
+}
+----
+
+Lets describe the fields:
```
moxious: Indicate which are mandatory.  My view is they should all be mandatory, but it's OK for example for properties to be an empty map {} but it cannot be null.

common/src/main/kotlin/streams/service/sink/strategy/CUDIngestionStrategy.kt
moxious: Also Lju brings up the point here -- what if labels = []?  You don't have to have labels if you're specifying an internal ID -- and this string concat may go wrong if labels is an empty array

common/src/main/kotlin/streams/service/sink/strategy/CUDIngestionStrategy.kt
conker84: We can give to the user the power to choose if DETACH or not. Wdyt?

Into the documentation there is an explanation of what the `detach` property means:
https://github.com/neo4j-contrib/neo4j-streams/pull/224/files#diff-96dcb9477dd8de76ceb7eb3db007e3dcR376

Do you think we can improve the description?

common/src/main/kotlin/streams/service/sink/strategy/CUDIngestionStrategy.kt
conker84: > you have to escape all identifiers and basically ALL inputs to this function with backticks. If I pass an identifier with a space in it, or a special character, this will break.
> 
> Double check code in getLabelsAsString for the same.

The `getLabelsAsString` method already escapes with the backticks the strings if necessary, I'll add a quote for identifiers

conker84: > Also Lju brings up the point here -- what if labels = []? You don't have to have labels if you're specifying an internal ID -- and this string concat may go wrong if labels is an empty array

Ok for this I think we should elaborate a little bit, in case of:
- Create: do we want to allow to create a node without a label?
- UPDATE/MERGE/DELETE: as Lju pointed out if we refer to the internal ID use the labels is an overhead, I'll remove this; but in other cases?

common/src/main/kotlin/streams/service/sink/strategy/CUDIngestionStrategy.kt
conker84: Ok I opened an issue for that, as it's a problem common to all strategies.
Look at #237 
Thanks!

common/src/test/kotlin/streams/service/sink/strategy/CUDIngestionStrategyTest.kt
conker84: Ok

common/src/test/kotlin/streams/service/sink/strategy/CUDIngestionStrategyTest.kt
conker84: Ok

common/src/test/kotlin/streams/service/sink/strategy/CUDIngestionStrategyTest.kt
conker84: This goes into #237

common/src/main/kotlin/streams/service/sink/strategy/CUDIngestionStrategy.kt
moxious: @conker84 labels are optional in Neo4j, so if we **don't** allow creating a node without label, we're going to get complaints and bug reports about this as a strange limitation of the sink.  So if it's not excessively complicated my view would be to allow it, and to print a warning or a suggestion that users probably shouldn't do this....we can also put in the docs that it's a bad idea.

To do the MATCH/MERGE, since indexes are always label-dependent, it's also going to be a terrible idea from a performance perspective.  Another possibility is to, with documentation, always turn `labels: []` into `labels: ["_Unlabeled"]` so that we can segregate and track these nodes, and always get a label index to be used.  In other words, for the labels array to have a default which is less than terrible -- and a warning to the user.

common/src/main/kotlin/streams/service/sink/strategy/CUDIngestionStrategy.kt
moxious: Making optional as you've done is a good choice, with default true.  I like it.  In terms of the description, I would just add a few words to summarize that detach means delete any incident relationships when you delete a node.   For Neo4j pros this is obvious, but some users will be lazy and not understand what "detach" means.  Link to the Cypher docs on detach delete so they can research further.

common/src/main/kotlin/streams/service/sink/strategy/CUDIngestionStrategy.kt
conker84: Yep no problem at all, I was just brainstorming :)

conker84: @moxious Let's manage `labels: []` and add an `N.B.` about the performance issues into the field description wdyt?

conker84:
@moxious it's ready

common/src/main/kotlin/streams/service/sink/strategy/CUDIngestionStrategy.kt
```diff
@@ -0,0 +1,271 @@
...
+                    it.value?.let {
+                        try {
+                            val data = toCUDEntity<CUDRelationship>(it)
+                            when (data?.op)  {
+                                CUDOperations.delete, null -> null // TODO send to the DLQ the null
+                                else -> if (data.from.ids.isNotEmpty() && data.to.ids.isNotEmpty() && data.properties.isNotEmpty()) data else null // TODO send to the DLQ the null
+                            }
+                        } catch (e: Exception) {
+                            null
```
jexp: what about reporting the actual exceptions to the error handler?

jexp: same below.

doc/asciidoc/consumer/index.adoc
```diff
@@ -313,6 +315,145 @@ By default no properties will be attached to the edge nodes but you can specify
...
+----
+streams.sink.topic.cud=<LIST_OF_TOPICS_SEPARATE_BY_SEMICOLON>
+----
+
+[source,ini]
+----
+streams.sink.topic.cud=my-topic;my-other.topic
+----
+
+We have one format for nodes:
```
jexp: where is that config set?

doc/asciidoc/consumer/index.adoc
```diff
@@ -313,6 +315,145 @@ By default no properties will be attached to the edge nodes but you can specify
 
 |===
 
+====== CUD File Format
+
+The CUD file format is JSON file that represents Graph Entities (Nodes/Relationships) and how to mange them in term
```
jexp: I'm not 100% sure how we manage the differentiation between this and the pattern based extraction?

PR #225, opened 2019-08-10
Docs Links Fixes

jexp:
Fixed links that are now broken on neo4j.com/docs/labs


moxious:
Thanks for doing this.  The docs worked -- but I think these links went bad in the switch-over to the new docs site which happened at the same time

PR #231, opened 2019-08-20
Improved DeadletterQueue, Logging and Error Handling

### LINKED ISSUE #222: Kafka Connect Sink - Handle the exceptions for non critical errors instead of throwing ConnectException
Sink Connector is throwing the connect exceptions when bad formatted event data is getting processed by templated cypher query.

However, due to the limitations of DLQ api not public for connect developers to move the bad messages to DLQ. Currently connector is throwing the ConnectException and going to failed state. We are not sure about when kafka connect team will enable this api(https://issues.apache.org/jira/browse/KAFKA-8597). Atleast can this connector handle the non critical exceptions and log the bad records and proceed with processing of next messages. May be some config parameter to opt for this default logging behavior verses DLQ. That way when KAFKA-8597 solution is available, then DLQ related functionaltiy can be added. This will enable the users to choose the config option for Logging vs DLQ depends on their requirements.

jexp:
Fixes #222 

- better parsing behavior for keys and values
- can also handle non-json keys with JSON serializer
- DLQ etc handling according to Kafka Connect, see 
  https://www.confluent.io/blog/kafka-connect-deep-dive-error-handling-dead-letter-queues


common/src/main/kotlin/streams/service/errors/KafkaDLQService.kt
```diff
@@ -0,0 +1,85 @@
...
+                    val props = Properties()
+                    props.putAll(config)
+                    props[ProducerConfig.KEY_SERIALIZER_CLASS_CONFIG] = ByteArraySerializer::class.java.name
+                    props[ProducerConfig.VALUE_SERIALIZER_CLASS_CONFIG] = ByteArraySerializer::class.java.name
+                    KafkaProducer<ByteArray,ByteArray>(props)
+                }
+    }
+
+    override fun report(errorDatas: List<ErrorData>) {
+        if (errorConfig.fail) throw ProcessingError(errorDatas)
```
moxious: The core conditional for what kind of error reporting to do is inside of the DLQ service.  But there might not be a DLQ.   So do I get this right the design we're going for is that you'll always have a DLQ service, just that if the topic isn't configured, the printing is done by the DLQ service?

This is OK but design wise seems somewhat strange; it seems like this might be better an "ErrorService" rather than a DLQ service.  Just one of its behaviors involves a DLQ.

I think the shape of this object is OK but it might not be named properly anymore.

consumer/src/main/kotlin/streams/kafka/KafkaAutoCommitEventConsumer.kt
```diff
@@ -91,18 +88,17 @@ open class KafkaAutoCommitEventConsumer(private val config: KafkaSinkConfigurati
...
             .let {
-                it.getOrDefault("error", emptyList<DLQData>())
-                        .forEach{ sentToDLQ(it as DLQData) }
+                it.get("error")?.let { handleErrors(it as List<ErrorData>) }
                 it.getOrDefault("ok", emptyList()) as List<StreamsSinkEntity>
             }
 
-    private fun sentToDLQ(dlqData: DLQData) {
-        dlqService?.send(config.streamsSinkConfiguration.dlqTopic, dlqData)
+    private fun handleErrors(errorData: List<ErrorData>) {
```
moxious: If all this function does is call a different method on errorService I don't see the point of it.  Just remove this function and inline the use of errorService.report

consumer/src/main/kotlin/streams/kafka/KafkaEventSink.kt
```diff
@@ -35,20 +36,12 @@ class KafkaEventSink(private val config: Config,
...
-                            .mapKeys { it.key.toString() }
-                            .toMutableMap()
-                    asProperties.remove(ConsumerConfig.KEY_DESERIALIZER_CLASS_CONFIG)
-                    asProperties.remove(ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG)
-                    KafkaDLQService(asProperties, "__streams.errors")
-                } else {
-                    null
-                }
+
+                val errorService = KafkaDLQService(kafkaConfig.asProperties(), ErrorService.ErrorConfig.from(kafkaConfig.streamsSinkConfiguration.errorConfig),{ s, e -> log.error(s,e as Throwable)})
```
moxious: Couple of comments here --
please break up this line, it's long, complicated, nested, I found it pretty tough to read.
Lastly that final argument:

```
{ s, e -> log.error(s,e as Throwable)
```

You initialize KafkaDLQService differently in another place, with a different final argument function (similar but still).  For consistency you might remove this argument altogether, or have it start with a default value so you can be consistent everywhere.   Within the StreamsEventConsumerFactory, I don't think it's this object's responsibility to specify error logging behavior.  Because separate code elsewhere deals with all of the config values, I think that's probably the right place to set this.

common/src/main/kotlin/streams/service/errors/ErrorService.kt
```diff
@@ -0,0 +1,118 @@
...
+            fun int(v:Any?) = when (v) {
+                null -> 0
+                is Int -> v
+                is String -> v.toInt()
+                else -> 0
+            }
+
+            fun from(config: Map<String, Any?>) =
+                    ErrorConfig(
+                            fail = config.getOrDefault(TOLERANCE, "none") == "none",
```
moxious: It's really, really easy for beginning users to screw up config when they're going fast and don't fully understand things.  This line has default behavior that fail is false unless tolerance = "none".  The actual valid values for tolerance are "all" or "none".  It'd be really nice to have a startup log message if the config value doesn't match the valid options -- this would really help debug things when users come in with problems, if the plugin tells you that it knows it's badly configured or configured in some way that has an effect the user doesn't expect.

common/src/main/kotlin/streams/service/errors/ErrorService.kt
```diff
@@ -0,0 +1,118 @@
...
+                            log = boolean(config.get(LOG)),
+                            logMessages = boolean(config.get(LOG_MESSAGES)),
+                            dlqTopic = config.get(DLQ_TOPIC) as String?,
```
moxious: What happens if dlqTopic is a string, but it's ""?   I'm pretty sure what happens is that delivery to the DLQ will hang and you get weird timeout errors in Kafka.   (That's what happens to me when I try to produce to something invalid/non-existant)

Also please do this tiny validation on a valid kafka topic name to ensure user didn't screw it up because often they will.  :)   

```
val legalChars = "[a-zA-Z0-9\\._\\-]"
private val maxNameLength = 255
private val rgx = new Regex(legalChars + "+")
```

Ref: https://stackoverflow.com/a/37067544/2920686

doc/asciidoc/consumer/index.adoc
```diff
@@ -315,47 +315,96 @@ By default no properties will be attached to the edge nodes but you can specify
...
+|===
+
+In Kafka Connect you can just configure the above properties.
 
-the Neo4j Streams plugin will re-route all the data to the the `my-dlq-topic` topic.
+For the Neo4j extension you prefix them with `streams.sink.` in the Neo4j configuration.
+
+Example settings:
+
+.Fail Fast, Abort
```
moxious: What does "abort" mean?  In your code, you throw an exception.  Then what?   Does that exception go to debug.log?  Do you crash neo4j?   Make a note of it.

doc/asciidoc/consumer/index.adoc
```diff
@@ -315,47 +315,96 @@ By default no properties will be attached to the edge nodes but you can specify
...
+[source]
+----
+errors.tolerance=none
+----
+
+
+.Ignore Errors, Log with Messages
```
moxious: Just call this "Log Errors with Messages".  You're not really ignoring them.  

Also you might note that you're logging batches, not individual messages.

kafka-connect-neo4j/docker/contrib.sink.string-json.neo4j.json
```diff
@@ -12,6 +12,6 @@
     "neo4j.authentication.basic.username": "neo4j",
     "neo4j.authentication.basic.password": "connect",
     "neo4j.encryption.enabled": false,
-    "neo4j.topic.cypher.my-topic": "MERGE (p:Person{name: event.name, surname: event.surname}) MERGE (f:Family{name: event.surname}) MERGE (p)-[:BELONGS_TO]->(f)"
+    "neo4j.topic.cypher.my-topic": "ME RGE (p:Person{name: event.name, surname: event.surname}) MERGE (f:Family{name: event.surname}) MERGE (p)-[:BELONGS_TO]->(f)"
```
moxious: beautiful.  I love this because this is a great example of what a user would actually screw up.  I did it myself a bunch of times.   Simple cypher errors in your config is (my most hated) error with this plugin.

kafka-connect-neo4j/src/main/kotlin/streams/kafka/connect/sink/Neo4jService.kt
```diff
@@ -144,4 +148,19 @@ class Neo4jService(private val config: Neo4jSinkConnectorConfig):
             throw ConnectException(exceptionMessages)
         }
     }
+
+    fun writeDataSync(data: Map<String, List<List<StreamsSinkEntity>>>) {
```
moxious: Your PR adds this function but I can't see that it's called anywhere.  I think this ties to the Tesla ticket and it's a good addition I want to keep, maybe I just missed how this gets referenced?

Need a config option that ties to the use of this rather than writeData

kafka-connect-neo4j/src/main/kotlin/streams/kafka/connect/sink/Neo4jSinkTask.kt
```diff
@@ -7,35 +7,48 @@ import org.apache.kafka.connect.sink.SinkRecord
...
     override fun version(): String {
         return VersionUtil.version(this.javaClass as Class<*>)
     }
 
     override fun start(map: Map<String, String>) {
         this.config = Neo4jSinkConnectorConfig(map)
+
+        val kafkaConfig = Properties()
+        kafkaConfig.putAll(map)
+        this.errorService = KafkaDLQService(kafkaConfig, ErrorService.ErrorConfig.from(kafkaConfig), log::error)
```
moxious: This is another instance of inconsistent third argument (log::error) to KafkaDLQService.  See earlier comment.

kafka-connect-neo4j/src/main/kotlin/streams/kafka/connect/sink/Neo4jSinkTask.kt
```diff
@@ -7,35 +7,48 @@ import org.apache.kafka.connect.sink.SinkRecord
...
-
-        neo4jService.writeData(data)
+        try {
+            val data = EventBuilder()
+                    .withBatchSize(config.batchSize)
+                    .withTopics(config.topics.allTopics())
+                    .withSinkRecords(collection)
+                    .build()
+
+            neo4jService.writeData(data)
```
moxious: See earlier comment, here we need some config switch to use writeDataSync if appropriate.

common/src/main/kotlin/streams/service/errors/ErrorService.kt
```diff
@@ -0,0 +1,118 @@
...
+                null -> 0
+                is Int -> v
+                is String -> v.toInt()
+                else -> 0
+            }
+
+            fun from(config: Map<String, Any?>) =
+                    ErrorConfig(
+                            fail = config.getOrDefault(TOLERANCE, "none") == "none",
+                            log = boolean(config.get(LOG)),
```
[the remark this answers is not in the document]
jexp: it's Map<String,Any> or Properties not a config object, (depending on where it's configured).

common/src/main/kotlin/streams/utils/JSONUtils.kt
```diff
@@ -112,12 +114,23 @@ object JSONUtils {
...
         val strValue = when (value) {
             is String -> value
+            is ByteArray -> String(value)
             else -> getObjectMapper().writeValueAsString(value)
         }
-        return getObjectMapper().readValue(strValue)
+        return try {
+            getObjectMapper().readValue(strValue)
+        } catch(e: JsonParseException) {
+            if (stringWhenFailure && String::class.isSubclassOf(T::class)) {
```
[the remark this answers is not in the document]
jexp: It's about the type of T which is the expected return type. Which can be Map or String or ByteArray or ...

It's about key handling when there are plain keys like `AB23` in a JSON, which jackson fails to decode, that's one of the issues CenturyLink has.

common/src/main/kotlin/streams/service/errors/KafkaDLQService.kt
jexp: Yes, true. I'll change that.

consumer/src/main/kotlin/streams/kafka/KafkaAutoCommitEventConsumer.kt
jexp: ya, it used to have more code.

consumer/src/main/kotlin/streams/kafka/KafkaEventSink.kt
jexp: no we need the argument for logging,
it's because different logger implementations (connect-plugin vs. neo4j extension) have different signatures, and here, unfortunately the signature cannot be resolved with  a method references as it's ambiguous.

doc/asciidoc/consumer/index.adoc
jexp: we log them individually: `errorDatas.forEach{log(it.toLogString(),it.exception)}`

kafka-connect-neo4j/docker/contrib.sink.string-json.neo4j.json
jexp: ups this was just for testing :) shouldn't have made it in :)

doc/asciidoc/consumer/index.adoc
jexp: fixed title

kafka-connect-neo4j/src/main/kotlin/streams/kafka/connect/sink/Neo4jService.kt
jexp: yep that was accidental, will be in another PR

kafka-connect-neo4j/src/main/kotlin/streams/kafka/connect/sink/Neo4jSinkTask.kt
jexp: see my earlier answer :)

kafka-connect-neo4j/src/main/kotlin/streams/kafka/connect/sink/Neo4jSinkTask.kt
jexp: yup, other PR

consumer/src/main/kotlin/streams/kafka/KafkaEventSink.kt
moxious: please leave a comment here on why then.  That makes sense but I would have missed that going through the code here, because the code looks trivial and refactorable

PR #233, opened 2019-08-21
warning about docker configuration in documentation

moxious:
This issue was found in debugging with @conker84 a while ago.   The issue was raised with Neo4j packaging, (@jennyowen) but it's not terribly simple for them to fix because the container is based on bash.

As a consequence of neo4j-streams using configuration that needs to end up in neo4j.conf, combined with Neo4j/Docker only allowing valid environment variables, this is just a limitation that needs to be documented to avoid user confusion.  Actual observed behavior with the plugin would be that if you used `topic-name` in the config, the Neo4j docker container would probably see it as just named "topic", and consumption would fail as that topic probably doesn't exist.

### DIFF
--- a/doc/asciidoc/docker/index.adoc
+++ b/doc/asciidoc/docker/index.adoc
@@ -8,6 +8,13 @@ This chapter describes Docker Compose templates that can be used to test Neo4j S
 --
 endif::env-docs[]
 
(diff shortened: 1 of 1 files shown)


conker84:
LGTM!

PR #257, opened 2019-10-18
fixes #252: Source with non-existant topic hangs indefinitely

### LINKED ISSUE #252: Source with non-existant topic hangs indefinitely
Will: If I configure Neo4j as a source for a topic that doesn’t exist in the kafka cluster, then any Cypher queries that would trigger publishing to this topic seem to hang. For example if I have

      streams.source.topic.nodes.fraudflags=Party{*}

but the topic I created in Kafka is fraud-flags instead of fraudflags then any Cypher queries that create or update a Party node just hang in Neo4j, presumable because the neo4j-streams plugin is unable to publish to a topic that doesn’t exist and needs to complete the publishing before the transaction can be closed. Is this the expected behavior? There was a DEBUG log message that neo4j-streams is trying to publish a transaction, which was enough to clue me in to investigate the cause of the hanging query but it might be nice to include which topic it is trying to publish to as well.

David: I’ve observed the same, but hadn’t looked into why it is yet.  I suspect what’s gonig on is that the client/transport layer accepts the message for delivery, and that what we’re perceiving as hanging is actually a message getting caught in a failure/retry loop within the Kafka driver.

conker84:
Fixes #252

This PR simply provides a discard mechanism for topics that are not defined in case the Kafka Broker property `auto.create.topics.enable` is set to `false`.

- For the Streams SINK, we just print the `invalid topics` because is not affect by the `block` problem
- For the Streams SOURCE:
  - we created the `KafkaAdminService` class that every `kafka.topic.discovery.polling.interval` checks for new topics into the Kafka Broker
  - the `KafkaEventRouter#sendEvent` method uses the `KafkaAdminService` in order to block the call to `Producer#send` method if the topic is not defined. N.b. in case `auto.create.topics.enable` is set to `true` the check return always `true`.

### DIFF
--- a/doc/asciidoc/producer/configuration.adoc
+++ b/doc/asciidoc/producer/configuration.adoc
@@ -15,6 +15,7 @@ kafka.connection.timeout.ms=10000
 kafka.replication=1
 kafka.linger.ms=1
 kafka.transactional.id=
+kafka.topic.discovery.polling.interval=300000
 
 streams.source.topic.nodes.<TOPIC_NAME>=<PATTERN>
 streams.source.topic.relationships.<TOPIC_NAME>=<PATTERN>
@@ -25,3 +26,10 @@ streams.source.schema.polling.interval=<MILLIS, the polling interval for getting
 Note: To use the Kafka transactions please set `kafka.transactional.id` and `kafka.acks` properly
 
 See the https://kafka.apache.org/documentation/#brokerconfigs[Apache Kafka documentation] for details on these settings.
+
+In case you Kafka broker is configured with `auto.create.topics.enable` to `false`,
+all the messages sent to topics that don't exist are discarded;
+this because the `KafkaProducer.send()` method blocks the execution, as explained in https://issues.apache.org/jira/browse/KAFKA-3539[KAFKA-3539].
+You can tune the custom property `kafka.topic.discovery.polling.interval` in order to
+periodically check for new topics into the Kafka cluster so the plugin will be able
+to send events to the defined topics.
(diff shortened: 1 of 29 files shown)


doc/asciidoc/producer/configuration.adoc
```diff
@@ -15,6 +15,7 @@ kafka.connection.timeout.ms=10000
 kafka.replication=1
 kafka.linger.ms=1
 kafka.transactional.id=
+kafka.topic.discovery.polling.interval=300000
```
moxious: This is a good start but request you add another bit of docs to a different section specifying how the plugin handles non-existant topics.  From reading the PR, the answer I guess is that it discards messages to non-existant topics, and warns at startup about this problem that it detects.  This polling interval controls how often we re-check for topics, so it can pick up on new topics created since the server started.

We just need to note this in the documentation, since otherwise to discover this plugin behavior you have to already know what the configuration item is, which people won't.

conker84: I extended the documentation with this:

_In case you Kafka broker is configured with `auto.create.topics.enable` to `false`,
all the messages sent to topics that don't exist are discarded;
this because the `KafkaProducer.send()` method blocks the execution, as explained in [KAFKA-3539](https://issues.apache.org/jira/browse/KAFKA-3539).
You can tune the custom property `kafka.topic.discovery.polling.interval` in order to
to send events to the defined topics._

moxious: @conker84 good to go as far as mergeability but I see the CI system is broken, so please investigate.  I'd merge on green build.

conker84: Something went wrong with Travis, I restarted it 3 times and it worked. Now it's green

PR #274, opened 2020-01-22
fixes #267: [Multidatabase] Source: Implement “from” syntax / config

### LINKED ISSUE #267: [Multidatabase] Source: Implement “from” syntax / config
Normally we would do this:

```
streams.source.topic.nodes.<TOPIC_NAME>=<PATTERN>
streams.source.topic.relationships.<TOPIC_NAME>=<PATTERN>
streams.source.enable=<true/false, default=true>
```

If the user configures nothing about databases we refer to the `default` database.

We can do the same fluent config trick by adding `from`.

```
streams.source.topic.nodes.<TOPIC_NAME>.from.<DB_NAME>=<PATTERN>
streams.source.topic.relationships.<TOPIC_NAME>.from.<DB_NAME>=<PATTERN>
streams.source.enable=<true/false, default=true>
```

conker84:
Fixes #267

This PR introduces the support for multi-tenancy for Neo4j 4.0 Enterprise Edition

  - added a new module `test-support` that is used only in tests for Source (and in the future sink). This module contains an extension of the `Neo4jContainter` for testing enterprise features; in this case in order to provide the test we need to ship the built jar to the container so I also added Maven Wrapper in order to build the project from Kotlin.
  - added the `from` syntax
  - split the `KafkaEventRouterIT` in the same way we did for the sink plugin in order to simplify the management 
  - updated the documentation


common/src/main/kotlin/streams/config/StreamsConfig.kt
```diff
@@ -78,4 +84,17 @@ class StreamsConfig(logService: LogService) : LifecycleAdapter() {
...
 
+    fun defaultDbName() = this.dbms.getDefaultDbName()
+
+    fun isDefaultDb(dbName: String) = this.defaultDbName() == dbName
+
+    fun isSourceGloballyEnabled() = this.config.getOrDefault(SOURCE_ENABLED, SOURCE_ENABLED_VALUE).toString().toBoolean()
+
+    fun isSourceEnabled(dbName: String) = this.config.getOrDefault("${SOURCE_ENABLED}.from.$dbName", isSourceGloballyEnabled()).toString().toBoolean()
+
+    fun hasProceduresGloballyEnabled() = this.config.getOrDefault(PROCEDURES_ENABLED, PROCEDURES_ENABLED_VALUE).toString().toBoolean()
```
moxious: this makes me realize that under 4.0 we should make a documentation note: when you enable procedures you do it globally for all databases.  This is OK for now but might be undesirable later.  If people use multi-tenant ops and some people want kafka in/out of database A, it might not make sense that procedures are enabled on database B.

What do you think about making this setting work like the rest - if you use it without a database name it works only for the default database, otherwise you have to enable it per-database?

common/src/main/kotlin/streams/config/StreamsConfigExtensionFactory.kt
```diff
@@ -15,6 +16,6 @@ class StreamsConfigExtensionFactory: ExtensionFactory<StreamsConfigExtensionFact
     }
 
     override fun newInstance(context: ExtensionContext, dependencies: Dependencies): Lifecycle {
-        return StreamsConfig(dependencies.log())
+        return StreamsConfig(dependencies.log(), dependencies.dbms())
```
moxious: I guess due to API changes this sort of thing needs to get backported to 3.5 because I think this API is a bit different right?

common/src/main/kotlin/streams/extensions/DatabaseManagementServiceExtensions.kt
```diff
@@ -1,7 +1,29 @@
...
 
-fun DatabaseManagementService.getSystemDb() = this.database(Neo4jUtils.SYSTEM_DATABASE_NAME) as GraphDatabaseAPI
\ No newline at end of file
+fun DatabaseManagementService.getSystemDb() = this.database(Neo4jUtils.SYSTEM_DATABASE_NAME) as GraphDatabaseAPI
+
+fun DatabaseManagementService.getDefaultDb() = this.database(Neo4jUtils.SYSTEM_DATABASE_NAME).let {
+    val defaultDbName = it.beginTx().use {
+        val columnAs = it.execute("SHOW DEFAULT DATABASE").columnAs<String>("name")
+        if (!columnAs.hasNext()) {
+            throw DatabaseNotFoundException("here is no default db into the instance")
```
moxious: This is a critical error that should never happen but will confuse a lot of people if it does.

Change error message to "Could not detect default database in this DBMS, or none is specified.  Please check your dbms.default_database and ensure that this database is online and available"

This system command you use is right, but things I'm not sure about is if it will respond with 1 record or 3 in CC, and also what happens if the DBMS knows what the default is, but it is offline or in error condition, and so on.

common/src/main/kotlin/streams/extensions/DatabaseManagementServiceExtensions.kt
```diff
@@ -1,7 +1,29 @@
...
+        columnAs.next()
+    }
+    this.database(defaultDbName) as GraphDatabaseAPI
+}
+
+fun DatabaseManagementService.getDefaultDbName() = this.database(Neo4jUtils.SYSTEM_DATABASE_NAME).let {
+    it.beginTx().use {
+        val columnAs = it.execute("SHOW DEFAULT DATABASE").columnAs<String>("name")
+        if (!columnAs.hasNext()) {
+            throw DatabaseNotFoundException("here is no default db into the instance")
```
moxious: same comment above

consumer/src/test/kotlin/integrations/kafka/KafkaEventSinkBase.kt
```diff
@@ -64,14 +64,15 @@ open class KafkaEventSinkBase {
     fun setUp() {
         graphDatabaseBuilder = ImpermanentDbmsRule()
                 .setConfig("kafka.bootstrap.servers", KafkaEventSinkSuiteIT.kafka.bootstrapServers)
-                .setConfig("kafka.zookeeper.connect", KafkaEventSinkSuiteIT.kafka.envMap["KAFKA_ZOOKEEPER_CONNECT"] ?: "")
+                .setConfig("kafka.zookeeper.connect", KafkaEventSinkSuiteIT.kafka.envMap["KAFKA_ZOOKEEPER_CONNECT"] ?: "localhost:2181" ?: "")
```
moxious: make sure to document what the default assumption is, otherwise this will be populated, users will get "can't connect message" and will file a ticket.  :)

doc/asciidoc/producer/configuration.adoc
```diff
@@ -32,4 +36,79 @@ all the messages sent to topics that don't exist are discarded;
...
+The old properties:
+
+----
+streams.source.enabled=<true/false, default=true>
+streams.source.topic.nodes.<TOPIC_NAME>=<PATTERN>
+streams.source.topic.relationships.<TOPIC_NAME>=<PATTERN>
+streams.procedures.enabled=<true/false, default=true>
+----
+
+are still valid and they refer to Neo4j's default db instance.
```
moxious: Add note: the default database is controlled by Neo4j's `dbms.default_database` configuration property so we're being clear about which default database applies for this user.

doc/asciidoc/producer/configuration.adoc
```diff
@@ -32,4 +36,79 @@ all the messages sent to topics that don't exist are discarded;
 this because the `KafkaProducer.send()` method blocks the execution, as explained in https://issues.apache.org/jira/browse/KAFKA-3539[KAFKA-3539].
 You can tune the custom property `kafka.topic.discovery.polling.interval` in order to
 periodically check for new topics into the Kafka cluster so the plugin will be able
-to send events to the defined topics.
\ No newline at end of file
+to send events to the defined topics.
+
+==== Multi Database Support
```
moxious: Great section here.  Please add answer to one question:

Will sink to / produce from fail if the database doesn't already exist?

Say I produce from foobar.  But I start a cold instance with nothing, and I want to create foobar later.  Does this work or does the whole thing die on startup?

This is not a requirement for the first version of 4.0 support but I bet people will ask for this, because we force peopel to pre-configure with neo4j.conf.  So imagine setting up a new machine.  It's a pain to start the cluster, create databases, shut down, reconfigure, restart, and then have your kafka.  It would be nice to configure just as you want, then start the cluster, then create the databases necessary.

Whichever way is the answer, just make it clear here in the docs

common/src/main/kotlin/streams/config/StreamsConfig.kt
conker84: Yes I agree with you, in fact, works already as you want because there is line 97:

https://github.com/neo4j-contrib/neo4j-streams/blob/38ee3379a5d6e6ca7825ce6198c8cf9b22e872e4/common/src/main/kotlin/streams/config/StreamsConfig.kt#L97

Which manages the procedure for each database instance.
If no value is specified for a database instance it inherits the configuration from the global configuration.
Please look at the `configuration.adoc` file for more details about it works

doc/asciidoc/producer/configuration.adoc
conker84: Great question, we can change the `StreamsConfig` to `DATABASE` extension so it will be loaded at runtime for each database instance, this should allow us to permit to the user to "update" the `neo4j.conf` file before creating a database and then load the new configuration for the instance. Wdyt?

moxious: that sounds workable.  :)

common/src/main/kotlin/streams/config/StreamsConfig.kt
moxious: great!  My miss.

common/src/main/kotlin/streams/config/StreamsConfigExtensionFactory.kt
conker84: Sorry, what do mean? This version is specific for Neo4j 4.0. We'll continue to maintain the version for Neo4j 3.5 in his branch.

common/src/main/kotlin/streams/extensions/DatabaseManagementServiceExtensions.kt
conker84: I'll remove the exception. and add the comment about `dbms.default_database` into the docs.
If it returns 3 records in CC I don't think is a problem we get the first this should be enough.
I don't think we should care about the state:
- `getDefaultDb` returns a `GraphDatabaseService` so we can use its APIs to check the status;
- `getDefaultDbName` returns a `String` so we don't care about the status;
Wdyt?

moxious: OK I agree, we don't care about the state in particular because also the state will change a lot over the operational lifetime of the database.   But do remember that since Kafka is generating async events into Neo4j, it's pretty much assured we're going to see errors that tie to this kind of scenario.  A kafka event comes in and as far as we know we should be able to process it, but because of reconciling state, we can't or it fails.  I guess this just then takes the regular error handling / DLQ path if these kinds of things come up.

common/src/main/kotlin/streams/config/StreamsConfigExtensionFactory.kt
moxious: Nevermind -- I got confused, this comment doesn't matter, you're right.

common/src/main/kotlin/streams/extensions/DatabaseManagementServiceExtensions.kt
conker84: Yes this is something that we need to manage with a hot reload function that reloads the whole plugin state.  Wdyt?

moxious: Sounds good....but keeping in mind that if the hot reload itself runs queries on the server, it needs throttling.  Imagine lots of messages coming in, all triggering errors, all triggering hot reloads.... :)

conker84:
Thanks a lot @moxious for the feedback!

PR #278, opened 2020-02-02
fixes #266: [Multidatabase] Sink: Implement “sink.to” syntax / config

### LINKED ISSUE #266: [Multidatabase] Sink: Implement “sink.to” syntax / config
Whenever we sink, we need to know which database that data is going into.

The default behaviour should be to go to the default database if the user doesn’t do any extra configuration.  

Example:

`streams.sink.topic.cypher.accountdetails=SOMECYPHER`

That goes to the default database.   Special note:  the “default default” database is called Neo4j, but we are **not** saying that this configuration should go to the neo4j database.  It should go to whatever is the system configured default database (which will usually be “neo4j”)

Suggestion -- introduce new config syntax `to.database` to make it look more fluent, and specify a DB.  Example:

`streams.sink.topic.cypher.accountdetails.to.myDatabase=SOMECYPHER`

In this form, we’ve explicitly said that the `accountdetails` topic should run cypher on “myDatabase” and the whole config key reads somewhat fluently.    Important:  This also raises the possibility that I could consume accountdetails on multiple different databases.  For example consider this:

```
streams.sink.topic.cypher.accountdetails.to.myDatabase=AccountDetails{*}
streams.sink.topic.cypher.accountdetails.to.anotherDatabase=MATCH (f:Whatever { id: event.id }) SET f.lastUpdate=datetime()
```

So the format is:

`streams.sink.topic.<strategy>.<topic>.to.<database>=whatever`

The full doc is here:
https://docs.google.com/document/d/1V2wqDE25C5XQtegMG4DaH4kmblxj_UHkVa-Iwp_6P8A

conker84:
Fixes #266

This PR implements the multi-tenancy support for both Neo4j and Kafka Connect plugins.

  - implemented the feature
  - updated documentation

### DIFF
--- a/common/src/main/kotlin/streams/config/StreamsConfig.kt
+++ b/common/src/main/kotlin/streams/config/StreamsConfig.kt
@@ -2,22 +2,22 @@ package streams.config
 
 import org.neo4j.dbms.api.DatabaseManagementService
 import org.neo4j.kernel.lifecycle.LifecycleAdapter
-import org.neo4j.logging.internal.LogService
+import org.neo4j.logging.Log
 import streams.extensions.getDefaultDbName
 import java.io.FileInputStream
 import java.io.FileNotFoundException
 import java.util.Properties
 import java.util.concurrent.ConcurrentHashMap
 
-class StreamsConfig(logService: LogService, private val dbms: DatabaseManagementService) : LifecycleAdapter() {
+data class StreamsConfig(private val log: Log, private val dbms: DatabaseManagementService) : LifecycleAdapter() {
 
     val config = ConcurrentHashMap<String, String>()
 
-    private val log = logService.getUserLog(StreamsConfig::class.java)
 
     private lateinit var neo4jConfFolder: String
 
     companion object {
+        private lateinit var INSTANCE: StreamsConfig
         private val SUPPORTED_PREFIXES = listOf("streams", "kafka")
         private const val SUN_JAVA_COMMAND = "sun.java.command"
         private const val CONF_DIR_ARG = "config-dir="
@@ -25,12 +25,16 @@ class StreamsConfig(logService: LogService, private val dbms: DatabaseManagement
         const val SOURCE_ENABLED_VALUE = true
         const val PROCEDURES_ENABLED = "streams.procedures.enabled"
         const val PROCEDURES_ENABLED_VALUE = true
+        const val SINK_ENABLED = "streams.sink.enabled"
+        const val SINK_ENABLED_VALUE = false
         const val DEFAULT_PATH = "."
         private var afterInitListeners = mutableListOf<((MutableMap<String, String>) -> Unit)>()
 
         fun registerListener(after: (MutableMap<String, String>) -> Unit) {
             afterInitListeners.add(after)
         }
+
+        fun getInstance() = INSTANCE
     }
 
     override fun init() {
@@ -40,6 +44,7 @@ class StreamsConfig(logService: LogService, private val dbms: DatabaseManagement
         neo4jConfFolder = getNeo4jConfFolder()
         loadConfiguration()
         afterInitListeners.forEach { it(config) }
+        INSTANCE = this
     }
 
     override fun stop() {
@@ -88,10 +93,10 @@ class StreamsConfig(logService: LogService, private val dbms: DatabaseManagement
 
         val properties = Properties()
         try {
-            log.info("the retrieved NEO4J_CONF dirs is $neo4jConfFolder")
+            log.info("The retrieved NEO4J_CONF dir is $neo4jConfFolder")
             properties.load(FileInputStream("$neo4jConfFolder/neo4j.conf"))
         } catch (e: FileNotFoundException) {
-            log.error("the neo4j.conf file is not under the directory defined into the directory $neo4jConfFolder, please set the NEO4J_CONF env correctly")
+            log.error("The neo4j.conf file is not under the directory defined into the directory $neo4jConfFolder, please set the NEO4J_CONF env correctly")
         }
         return properties
     }
@@ -124,4 +129,9 @@ class StreamsConfig(logService: LogService, private val dbms: DatabaseManagement
     fun hasProceduresGloballyEnabled() = this.config.getOrDefault(PROCEDURES_ENABLED, PROCEDURES_ENABLED_VALUE).toString().toBoolean()
 
     fun hasProceduresEnabled(dbName: String) = this.config.getOrDefault("${PROCEDURES_ENABLED}.$dbName", hasProceduresGloballyEnabled()).toString().toBoolean()
+
+    fun isSinkGloballyEnabled() = this.config.getOrDefault(SINK_ENABLED, SINK_ENABLED_VALUE).toString().toBoolean()
+
+    fun isSinkEnabled(dbName: String) = this.config.getOrDefault("${SINK_ENABLED}.to.$dbName", isSinkGloballyEnabled()).toString().toBoolean()
+
 }
--- a/common/src/main/kotlin/streams/service/Topics.kt
+++ b/common/src/main/kotlin/streams/service/Topics.kt
@@ -3,6 +3,13 @@ package streams.service
 import streams.service.sink.strategy.*
 import kotlin.reflect.jvm.javaType
 
+class TopicValidationException(message: String): RuntimeException(message)
+
+private fun TopicType.replaceKeyBy(replacePrefix: Pair<String, String>) = if (replacePrefix.first.isNullOrBlank())
+        this.key
+    else
+        this.key.replace(replacePrefix.first, replacePrefix.second)
+
 data class Topics(val cypherTopics: Map<String, String> = emptyMap(),
                   val cdcSourceIdTopics: Set<String> = emptySet(),
                   val cdcSchemaTopics: Set<String> = emptySet(),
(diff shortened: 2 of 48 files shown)


conker84:
@moxious I created this as draft because we need to close before the `Source` PR.

common/src/main/kotlin/streams/config/StreamsConfig.kt
```diff
@@ -124,4 +129,9 @@ class StreamsConfig(logService: LogService, private val dbms: DatabaseManagement
     fun hasProceduresGloballyEnabled() = this.config.getOrDefault(PROCEDURES_ENABLED, PROCEDURES_ENABLED_VALUE).toString().toBoolean()
 
     fun hasProceduresEnabled(dbName: String) = this.config.getOrDefault("${PROCEDURES_ENABLED}.$dbName", hasProceduresGloballyEnabled()).toString().toBoolean()
+
+    fun isSinkGloballyEnabled() = this.config.getOrDefault(SINK_ENABLED, SINK_ENABLED_VALUE).toString().toBoolean()
+
+    fun isSinkEnabled(dbName: String) = this.config.getOrDefault("${SINK_ENABLED}.to.$dbName", isSinkGloballyEnabled()).toString().toBoolean()
```
moxious: point of documentation.  If you have:

```
streams.sink.enabled=true
streams.sink.enabled.to.foo=false
```

Then sink is enabled on all databases EXCEPT foo.  (Local overrides global)

common/src/main/kotlin/streams/service/Topics.kt
```diff
@@ -26,13 +38,16 @@ data class Topics(val cypherTopics: Map<String, String> = emptyMap(),
...
-            val sourceIdKey = TopicType.CDC_SOURCE_ID.key.replace(prefix, toReplace)
-            val schemaKey = TopicType.CDC_SCHEMA.key.replace(prefix, toReplace)
-            val cudKey = TopicType.CUD.key.replace(prefix, toReplace)
-            val nodePatterKey = TopicType.PATTERN_NODE.key.replace(prefix, toReplace)
-            val relPatterKey = TopicType.PATTERN_RELATIONSHIP.key.replace(prefix, toReplace)
+        fun from(map: Map<String, Any?>, replacePrefix: Pair<String, String> = ("" to ""), dbName: String = "", invalidTopics: List<String> = emptyList()): Topics {
+            val config = map
+                    .filterKeys { if (dbName.isNotBlank()) it.endsWith(".to.$dbName") else !it.contains(".to.") }
+                    .mapKeys { if (dbName.isNotBlank()) it.key.replace(".to.$dbName", "") else it.key }
+            val cypherTopicPrefix = if (replacePrefix.first.isNullOrBlank()) TopicType.CYPHER.key else TopicType.CYPHER.key.replace(replacePrefix.first, replacePrefix.second)
```
moxious: these if/elses are hard to read because of so much repeated code.  Define a helper function that does ```if(replacePrefix.first.isNullOrBlank()) INPUT.key else INPUT.key.replace(replacePrefix.first, replacePrefix.second)```

It'd be very easy for there to be a typo here we would miss

consumer/src/main/kotlin/streams/kafka/KafkaEventSink.kt
```diff
@@ -75,19 +77,20 @@ class KafkaEventSink(private val config: Map<String, String>,
...
 
     override fun start() { // TODO move to the abstract class
-        val streamsConfig = StreamsSinkConfiguration.from(config)
+        val streamsConfig = StreamsSinkConfiguration.from(config, db.databaseName())
         val topics = streamsTopicService.getTopics()
         val isWriteableInstance = Neo4jUtils.isWriteableInstance(db)
         if (!streamsConfig.enabled) {
             if (topics.isNotEmpty() && isWriteableInstance) {
-                log.warn("You configured the following topics: $topics, in order to make the Sink work please set the `${StreamsSinkConfigurationConstants.STREAMS_CONFIG_PREFIX}${StreamsSinkConfigurationConstants.ENABLED}` to `true`")
+                log.warn("You configured the following topics: $topics, in order to make the Sink work please set the `${StreamsConfig.SINK_ENABLED}` to `true`")
```
moxious: make it copy/pasteable.   "Please set ${StreamsConfig.SINK_ENABLED}=true"

consumer/src/main/kotlin/streams/kafka/KafkaEventSink.kt
```diff
@@ -108,6 +115,7 @@ class KafkaEventSink(private val config: Map<String, String>,
         StreamsUtils.ignoreExceptions({
             runBlocking {
                 if (job.isActive) {
+                    eventConsumer.wakeup()
```
moxious: I know we're not concerned with WakeupException here based on what it means.  But if it happens -- then job.cancelAndJoin() doesn't happen.  Is that OK?  Watch out for that possibility.

consumer/src/test/kotlin/streams/StreamsSinkConfigurationTest.kt
```diff
@@ -1,14 +1,79 @@
...
+        val config = mapOf("streams.sink.polling.interval" to pollingInterval,
+                topicKey to topicValue,
+                topicKeyNeo to topicValueNeo,
+                topicKeyFoo to topicValueFoo)
+        streamsConfig.config.putAll(config)
+        val streamsSinkConf = StreamsSinkConfiguration.from(streamsConfig, defalutDbName)
+        val cypherTopics = streamsSinkConf.topics.asMap()[TopicType.CYPHER] as Map<String, String>
+        assertEquals(mapOf("myTopic" to topicValue, "myTopicNeo" to topicValueNeo), cypherTopics)
+    }
+
```
moxious: Add test -- attempt to do this should fail:

streams.sink.topic.cypher.whatever.to.system=whatever

Should always fail.   "system" is not a valid sink target.

doc/asciidoc/consumer/configuration.adoc
```diff
@@ -20,4 +20,85 @@ kafka.value.deserializer=org.apache.kafka.common.serialization.ByteArrayDeserial
...
+----
+streams.sink.topic.cypher.<TOPIC_NAME>.to.<DB_NAME>=<CYPHER_QUERY>
+streams.sink.topic.cdc.sourceId.to.<DB_NAME>=<LIST_OF_TOPICS_SEPARATE_BY_SEMICOLON>
+streams.sink.topic.cdc.schema.to.<DB_NAME>=<LIST_OF_TOPICS_SEPARATE_BY_SEMICOLON>
+streams.sink.topic.pattern.node.<TOPIC_NAME>.to.<DB_NAME>=<NODE_EXTRACTION_PATTERN>
+streams.sink.topic.pattern.relationship.<TOPIC_NAME>.to.<DB_NAME>=<RELATIONSHIP_EXTRACTION_PATTERN>
+streams.sink.enabled.to.<DB_NAME>=<true/false, default=true>
+streams.procedures.enabled.<DB_NAME>=<true/false, default=true>
+----
+
```
moxious: Add:  you may not specify system as DB_NAME.   Database names are case-sensitive, and must follow Neo4j database naming rules (Reference: https://neo4j.com/docs/operations-manual/current/manage-databases/configuration/#manage-databases-administration)

doc/asciidoc/consumer/configuration.adoc
```diff
@@ -20,4 +20,85 @@ kafka.value.deserializer=org.apache.kafka.common.serialization.ByteArrayDeserial
...
+----
+streams.sink.topic.cypher.<TOPIC_NAME>=<CYPHER_QUERY>
+streams.sink.topic.cdc.sourceId=<LIST_OF_TOPICS_SEPARATE_BY_SEMICOLON>
+streams.sink.topic.cdc.schema=<LIST_OF_TOPICS_SEPARATE_BY_SEMICOLON>
+streams.sink.topic.pattern.node.<TOPIC_NAME>=<NODE_EXTRACTION_PATTERN>
+streams.sink.topic.pattern.relationship.<TOPIC_NAME>=<RELATIONSHIP_EXTRACTION_PATTERN>
+streams.sink.enabled=<true/false, default=true>
+----
+
+are still valid and they refer to Neo4j's default db instance.
```
moxious: ...which is usually called "neo4j", but can be controlled by separate Neo4j system configuration.

common/src/main/kotlin/streams/config/StreamsConfig.kt
conker84: Done

common/src/main/kotlin/streams/service/Topics.kt
conker84: Done

consumer/src/main/kotlin/streams/kafka/KafkaEventSink.kt
conker84: Done

consumer/src/main/kotlin/streams/kafka/KafkaEventSink.kt
conker84: The Kafka documentation states:
```
/**
  * Wakeup the consumer. This method is thread-safe and is useful in particular to abort a long poll.
  * The thread which is blocking in an operation will throw {@link org.apache.kafka.common.errors.WakeupException}.
  * If no thread is blocking in a method which can throw {@link org.apache.kafka.common.errors.WakeupException}, the next call to such a method will raise it instead.
  */
```

So the `WakeupException` should be thrown into the coroutine, here:

https://github.com/neo4j-contrib/neo4j-streams/blob/56e99d5bd38851020c457b15a99cec7319cd3e48/consumer/src/main/kotlin/streams/kafka/KafkaEventSink.kt#L149-L155

consumer/src/test/kotlin/streams/StreamsSinkConfigurationTest.kt
conker84: We block the creation of the sink/source plugins at `Extension` level:

https://github.com/neo4j-contrib/neo4j-streams/blob/e17d6d64d33a8d1ae61d8a07f6a6193a06f2c551/producer/src/main/kotlin/streams/StreamsExtensionFactory.kt#L50-L52

and

https://github.com/neo4j-contrib/neo4j-streams/blob/e17d6d64d33a8d1ae61d8a07f6a6193a06f2c551/consumer/src/main/kotlin/streams/StreamsEventSinkExtensionFactory.kt#L45-L47

Do you want to add a check also here?

doc/asciidoc/consumer/configuration.adoc
conker84: Done

doc/asciidoc/consumer/configuration.adoc
conker84: Done

consumer/src/test/kotlin/streams/StreamsSinkConfigurationTest.kt
moxious: Sorry if I missed it, if the check is already in there it isn't double-needed

PR #281, opened 2020-02-12
apply Apache 2.0 license to repository

moxious:
Applies Apache 2.0 licensing to the repository, as discussed with Andrea & Michael.


moxious:
The actual license copy is a direct carry-over from the APOC repo

PR #299, opened 2020-04-03
Fixes #293: streams.sink.polling.interval is possibly confusing and m…

### LINKED ISSUE #293: streams.sink.polling.interval is possibly confusing and may not be used anyway
Here are all references to this config key in the source:

https://github.com/neo4j-contrib/neo4j-streams/search?q=sink.polling.interval&unscoped_q=sink.polling.interval

I can't see where it's used anywhere.   It also sorta doesn't make sense that this property key would do anything, because it's not SCHEMA polling, and if it's a kafka client setting, then it should be kafka.polling.interval.

So I think this key should maybe be removed from docs & code, because it creates confusion with schema polling, and also with kafka client settings like kafka.max.poll.interval.ms

mroiter-larus:
Fixes #293 

Removed all references to `streams.sink.polling.interval`


conker84:
@mroiter-larus thank you!

PR #309, opened 2020-05-11
Fixes #305: CDC and Kafka Log Compaction

### LINKED ISSUE #305: CDC and Kafka Log Compaction
Hi everyone !! This issue is somehow similar to this one  #272 but the focus here is on Neo4j Streams Source for CDC in tandem with Kafka Log Compaction rather than on Neo4j Streams Procedures.

When using Change Data Capture with Kafka, one of the most common use cases is managing Kafka partitions as logs. In this scenario, Log Compaction is somehow necessary for CDC: we need a complete dataset but we must also avoid unlimited growth in the size of partitions. From the Kafka documentation:

> **Log compaction** ensures that Kafka will always retain at least the last known value for each message key within the log of data for a single topic partition. It **addresses use cases and scenarios such as restoring state after application crashes or system failure or reloading caches after application restarts during operational maintenance.**

On data retention:

> So far we have described only the simpler approach to data retention where old log data is discarded after a fixed period or when the log reaches some predetermined size. **This works well for temporal event data** such as logging **where each record stands alone. However, an important class of data streams are the log of changes to keyed, mutable data (for example, the changes to a database table).**

The current implementation of Neo4j Streams prevents this use cases since the message key is a combination of the transaction id and the id of the event inside the transaction, as shown in the following code ([KafkaEventRouter.kt](https://github.com/neo4j-contrib/neo4j-streams/blob/3.5/producer/src/main/kotlin/streams/kafka/KafkaEventRouter.kt)- line 86 - **sendEvent**  method):
```
private fun sendEvent(partition: Int, topic: String, event: StreamsTransactionEvent) {
        if (log.isDebugEnabled) {
            log.debug("Trying to send a transaction event with txId ${event.meta.txId} and txEventId ${event.meta.txEventId} to kafka")
        }
        val producerRecord = ProducerRecord(topic, partition, System.currentTimeMillis(), 
                 "${event.meta.txId + event.meta.txEventId}-${event.meta.txEventId}",
        send(producerRecord)
    }
```
As far as I know, this will generate a sequence of unique keys, preventing Log Compaction. Instead, choosing the node or relationship id as the message key (``"${event.payload.id}"``) would enable Log Compaction on the Kafka side. However, I'm pretty sure there are specific reasons for creating the key that way: maybe the message keys are used to know where we are in the Neo4j log, to recover after a failure, but I'm not sure.

Can you explain why the keys are generated in this way? Is it possible to enable Log Compaction in Neo4j Streams right now? If the answer is no, I think this is a must-have feature for Change Data Capture. Thanks !!

## Considered alternatives

While writing this issue, I actually came up with a possible alternative solution (the same also applies to the issue #304): one can simply create a microservice using Kafka Streams to add custom logic and cope with order guarantees and log compaction. However, I am still convinced that these problems should be addressed within Neo4j Streams.

mroiter-larus:
Fixes #305

### DIFF
--- a/common/src/main/kotlin/streams/utils/KafkaValidationUtils.kt
+++ b/common/src/main/kotlin/streams/utils/KafkaValidationUtils.kt
@@ -34,4 +34,16 @@ object KafkaValidationUtils {
                 ?.value()
                 ?.toBoolean() ?: false
     }
+
+    fun isLogCompactionEnabled(kafkaProps: Properties) = isLogCompactionEnabled(AdminClient.create(kafkaProps))
+
+    fun isLogCompactionEnabled(client: AdminClient): Boolean {
+        val firstNodeId = client.describeCluster().nodes().get().first().id()
+        val configs = client.describeConfigs(listOf(ConfigResource(ConfigResource.Type.BROKER, firstNodeId.toString()))).all().get()
+        return configs.values
+                .flatMap { it.entries() }
+                .find { it.name() == "log.cleanup.policy" }
+                ?.value()
+                ?.toString().equals("compact")
+    }
 }
--- a/producer/src/main/kotlin/streams/kafka/KafkaAdminService.kt
+++ b/producer/src/main/kotlin/streams/kafka/KafkaAdminService.kt
@@ -49,4 +49,6 @@ class KafkaAdminService(private val props: KafkaConfiguration, private val allTo
     }
 
     fun getInvalidTopics() = KafkaValidationUtils.getInvalidTopics(client, allTopics)
+
+    fun isLogCompactionEnabled() = KafkaValidationUtils.isLogCompactionEnabled(client)
 }
--- a/producer/src/main/kotlin/streams/kafka/KafkaEventRouter.kt
+++ b/producer/src/main/kotlin/streams/kafka/KafkaEventRouter.kt
@@ -91,9 +91,17 @@ class KafkaEventRouter: StreamsEventRouter {
         if (log.isDebugEnabled) {
             log.debug("Trying to send a transaction event with txId ${event.meta.txId} and txEventId ${event.meta.txEventId} to kafka")
         }
-        val producerRecord = ProducerRecord(topic, partition, System.currentTimeMillis(), "${event.meta.txId + event.meta.txEventId}-${event.meta.txEventId}",
-                JSONUtils.writeValueAsBytes(event))
-        send(producerRecord)
+        if (kafkaAdminService.isLogCompactionEnabled()) {
+            val producerRecord = ProducerRecord(topic, partition, System.currentTimeMillis(),
+                    "${event.payload.id}",
+                    JSONUtils.writeValueAsBytes(event))
+            send(producerRecord)
+        }
+        else {
+            val producerRecord = ProducerRecord(topic, partition, System.currentTimeMillis(), "${event.meta.txId + event.meta.txEventId}-${event.meta.txEventId}",
+                    JSONUtils.writeValueAsBytes(event))
+            send(producerRecord)
+        }
     }
 
     override fun sendEvents(topic: String, transactionEvents: List<out StreamsEvent>) {
--- a/producer/src/test/kotlin/streams/integrations/KafkaEventRouterLogCompactionBaseTSE.kt
+++ b/producer/src/test/kotlin/streams/integrations/KafkaEventRouterLogCompactionBaseTSE.kt
@@ -1,0 +1,52 @@
+package streams.integrations
+
+import org.apache.kafka.clients.consumer.KafkaConsumer
+import org.junit.After
+import org.junit.AfterClass
+import org.junit.Before
+import org.junit.BeforeClass
+import org.neo4j.test.rule.DbmsRule
+import org.neo4j.test.rule.ImpermanentDbmsRule
+import streams.KafkaTestUtils
+import streams.setConfig
+
+open class KafkaEventRouterLogCompactionBaseTSE { // TSE (Test Suite Element)
+
+    companion object {
+
+        private var startedFromSuite = true
+
+        @BeforeClass
+        @JvmStatic
+        fun setUpContainer() {
+            if (!KafkaEventRouterLogCompactionSuiteIT.isRunning) {
+                startedFromSuite = false
+                KafkaEventRouterLogCompactionSuiteIT.setUpContainer()
+            }
+        }
+
+        @AfterClass
+        @JvmStatic
+        fun tearDownContainer() {
+            if (!startedFromSuite) {
+                KafkaEventRouterLogCompactionSuiteIT.tearDownContainer()
+            }
+        }
+    }
+
+    val db: DbmsRule = ImpermanentDbmsRule()
+
+    lateinit var kafkaConsumer: KafkaConsumer<String, ByteArray>
+
+    @Before
+    fun setUp() {
+        kafkaConsumer = KafkaTestUtils.createConsumer(bootstrapServers = KafkaEventRouterLogCompactionSuiteIT.kafka.bootstrapServers)
+        db.setConfig("kafka.bootstrap.servers", KafkaEventRouterLogCompactionSuiteIT.kafka.bootstrapServers)
+    }
+
+    @After
+    fun tearDown() {
+        db.shutdown()
+        kafkaConsumer.close()
+    }
+}
--- a/producer/src/test/kotlin/streams/integrations/KafkaEventRouterLogCompactionSuiteIT.kt
+++ b/producer/src/test/kotlin/streams/integrations/KafkaEventRouterLogCompactionSuiteIT.kt
@@ -1,0 +1,56 @@
+package streams.integrations
+
+import org.junit.AfterClass
+import org.junit.Assume
+import org.junit.BeforeClass
+import org.junit.runner.RunWith
+import org.junit.runners.Suite
+import org.testcontainers.containers.KafkaContainer
+import streams.utils.StreamsUtils
+
+@RunWith(Suite::class)
+@Suite.SuiteClasses(
+        KafkaEventRouterLogCompactionTSE::class
+)
+class KafkaEventRouterLogCompactionSuiteIT {
+
+    companion object {
+        /**
+         * Kafka TestContainers uses Confluent OSS images.
+         * We need to keep in mind which is the right Confluent Platform version for the Kafka version this project uses
+         *
+         * Confluent Platform | Apache Kafka
+         *                    |
+         * 4.0.x	          | 1.0.x
+         * 4.1.x	          | 1.1.x
+         * 5.0.x	          | 2.0.x
+         *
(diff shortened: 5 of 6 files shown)


moxious:
I have a lot of thoughts here.   Pinging @Solom00n here as the original issue author so they can follow along.

First -- to change the keying strategy depending on the topic compaction configuration is too magical.  Also makes me worry:  can the compaction strategy change after the topic is created?  This raises the possibility that the keying strategy on the stream can change between system restarts if the upstream broker config is detected to have changed.  It seems more appropriate to be able to select a keying strategy via neo4j-streams config, if one is wanted.

Second -- it's not clear to me that there is a better keying strategy here.  By keying the messages by TXID, we're essentially "externalizing the transaction logs" of neo4j to kafka.  The TXID is the natural thing that neo4j itself gives to neo4j-streams to do.

If this were a relational database, it'd be better to key messages with something like a table name and a PK_ID.  This would enable log compaction to compact down to the actual current state of the table in a sense.  But it isn't a relational database; it would seem that we'd need to key to a label combination + PK per node, and reltype + PK per relationship.   Trouble with that is that the TXs that we're externalizing are a combination of many of those.   To make this kind of log compaction work in the way that would be best for kafka, it would seem that 1 Neo4j TX needs to get split into n different keyed messages sent upstream.  (One per node/rel affected).  Then of course we'd need to deal with situations like, "What do you do if a TX involved the need for 2k messages, and 15 of them couldn't be sent?"   (The whole/part problem)

I see in this PR that we've got an alternative keying strategy, but I'm not sure that @mroiter-larus 's new keying strategy ("${event.payload.id}") addresses the original issue.   To address the original issue it seems to me like the TX payload needs to get blown up into a lot of different messages, but this too has problems.

conker84:
> I have a lot of thoughts here. Pinging @Solom00n here as the original issue author so they can follow along.
> 
> First -- to change the keying strategy depending on the topic compaction configuration is too magical. Also makes me worry: can the compaction strategy change after the topic is created? This raises the possibility that the keying strategy on the stream can change between system restarts if the upstream broker config is detected to have changed. It seems more appropriate to be able to select a keying strategy via neo4j-streams config, if one is wanted.

IHMO I agree with you I think that a configuration param could be the best solution.

> Second -- it's not clear to me that there is a better keying strategy here. By keying the messages by TXID, we're essentially "externalizing the transaction logs" of neo4j to kafka. The TXID is the natural thing that neo4j itself gives to neo4j-streams to do.
> 
> If this were a relational database, it'd be better to key messages with something like a table name and a PK_ID. This would enable log compaction to compact down to the actual current state of the table in a sense. But it isn't a relational database; it would seem that we'd need to key to a label combination + PK per node, and reltype + PK per relationship. Trouble with that is that the TXs that we're externalizing are a combination of many of those. To make this kind of log compaction work in the way that would be best for kafka, it would seem that 1 Neo4j TX needs to get split into n different keyed messages sent upstream. (One per node/rel affected). Then of course we'd need to deal with situations like, "What do you do if a TX involved the need for 2k messages, and 15 of them couldn't be sent?" (The whole/part problem)

I think that we should evaluate a more general solution because labels are mutable so my suggestion is to use:

* `node-<internal_node_id>`
* `rel-<internal_rel_id>`

and this should be a general solution that addresses the @Solom00n 

> I see in this PR that we've got an alternative keying strategy, but I'm not sure that @mroiter-larus 's new keying strategy ("${event.payload.id}") addresses the original issue. To address the original issue it seems to me like the TX payload needs to get blown up into a lot of different messages, but this too has problems.

Please let me know what are your thoughts about!

moxious:
so supposing we published node-ID and rel-ID -- this does still require breaking a single TX up into a lot of different individual changes, right?  And these changes need to be sequenced carefully, right?  (For example, it would not make sense to delete a node before its incident relationships, since a partial view could have dangling relationships).  What could we put in place to minimize or eliminate the risk of a partial transaction being sent to Kafka?

Solom00n:
Hi @moxious !! Sorry but I don't get your point. The CDC stream produced by Neo4j Stream should reflect already **committed** transactions so there should be no risk of a partial transaction being sent to Kafka. Moreover, to enable log compaction, you don't need to change the currently produced stream of events (which I assume is consistent) but simply send them to Kafka with a different message key. 

Can you expand your thoughts a little bit more, making some concrete example of a hypothetical problematic situation?

> I see in this PR that we've got an alternative keying strategy, but I'm not sure that @mroiter-larus 's new keying strategy ("${event.payload.id}") addresses the original issue.

@conker84, indeed this new keying strategy is not enough: we are not considering the compaction of deleted entities. From the Kafka documentation ([https://kafka.apache.org/documentation/#compaction](https://kafka.apache.org/documentation/#compaction)):
> Log compaction is a mechanism to give finer-grained per-record retention, rather than the coarser-grained time-based retention. **The idea is to selectively remove records where we have a more recent update with the same primary key. This way the log is guaranteed to have at least the last state for each key** ... Compaction also allows for deletes. **A message with a key and a null payload will be treated as a delete from the log**. This delete marker will cause any prior message with that key to be removed (as would any new message with that key) ...

Let's say that we have an event for an entity whose key is X and see what happens if we use ``${event.payload.id}`` as the message key:
- **Create event:** the only problem here is that Neo4j reuses its internal ids when nodes and relationships are deleted (see [https://community.neo4j.com/t/is-it-reliable-to-depend-on-node-id-to-query-data/10299](https://community.neo4j.com/t/is-it-reliable-to-depend-on-node-id-to-query-data/10299)) therefore, in general, It's not recommended to use the internal id. However, this shouldn't be a real problem if we consider that:
  1. if X has been used in the past, we already have a delete marker in the Kafka log so any prior message with an X key will be removed. Moreover, any new message with an X key (e.g. our create message) will cause any prior message with that key to be removed. In the end, what we get reading the log from the beginning, is precisely our newly created entity.
  2. if there is no previous message with an X key, it means that they have already been compacted away, so no problem. 

- **Update event:** records with an X key are selectively removed if we have a more recent update with the same X key

- **-->Delete event<--:** to enable compaction we must send to Kafka a message with a null payload. One possible solution here is to send a regular delete message with an X key followed by an X:null message, which acts as a tombstone enabling the compaction of the log.

**Note:** if you need to insert more information in the key, you may consider the possibility of using a more complex data structure instead of a simple string or other primitive types.

Of course, all the above scenarios should be tested, but it seems to me that using the internal id of an entity is enough to enable log compaction. Let me know what you think of all this.

moxious:
Concrete example, say we do this in Neo4j:

```
CREATE (p:Person { name: "Bob"}) CREATE (p2:Person { name: "Susan" }) CREATE (p3:Person { name: "Bill" }) CREATE (p)-[:KNOWS]->(p3)
```

That's *1* transaction consisting of 3 nodes and 1 relationship.

If we key by rel-ID and node-ID, then we need to send 4 messages to kafka to comprise this 1 transaction.  Let's call these messages A, B, C, D (this final one is the relationship, for arguments sake)

One message per unique rel-ID / node-ID.   If we did this, then via log compaction, the logs would bring the data up to the current content of the database with time, rather than a full log of every mutation that ever occurred anywhere.

Also if we do this, we have to deal with the possibility, what if we send A, B, C, D and there's an error with B & C, and only A and D go through?  The result is nonsense -- a dangling relationship that can't exist, and missing nodes.   This "partial problem" tends to argue in favor of sending the entire TX.  But if you do *that*, sensible keying that enables log compaction is basically impossible

Solom00n:
> One message per unique rel-ID / node-ID. If we did this, then via log compaction, the logs would bring the data up to the current content of the database with time, rather than a full log of every mutation that ever occurred anywhere.

Yes, that's right !! You will not have a full log of every mutation for each entity since the beginning of time. Maintaining such a log is not scalable for long-running distributed application since it's size will grow without a limit. Moreover, from a user perspective, I don't need a full log: what I need is a sequence of events that let me reconstruct the current state of the database (more probably, a part of it), for example, to cope with failure and subsequent restart of one of my service.

> Also if we do this, we have to deal with the possibility, what if we send A, B, C, D and there's an error with B & C, and only A and D go through? The result is nonsense -- a dangling relationship that can't exist, and missing nodes. This "partial problem" tends to argue in favor of sending the entire TX. But if you do that, sensible keying that enables log compaction is basically impossible

Now I see !! This is actually a problem but there is a simple solution: it's possible to use transactional semantics, which enables atomic write:
> Transactions enable atomic writes to multiple Kafka topics and partitions. **All of the messages included in the transaction will be successfully written or none of them will be.** For example, an error during processing can cause a transaction to be aborted, in which case none of the messages from the transaction will be readable by consumers.

See [https://www.confluent.io/blog/transactions-apache-kafka/](https://www.confluent.io/blog/transactions-apache-kafka/). Using a Kafka transaction for each Neo4j transaction there should be no "partial problem". Let me know if you think It's a step in the right direction.

moxious:
@Solom00n with transactional semantics it seems completely doable.   

However this is also a bigger change.   @mroiter-larus and @conker84 here's how I would now see this.  Just as our plugin has multiple "sink strategies", what we're really discussing here isn't to do with keying, it's really a second "source strategy".   We can't necessarily change it just on the basis of broker config and actually meet the original need.  Because if we do that -- the broker config would fundamentally change the publishing format which would be quite weird.

My thinking now is that we need to maintain the default publishing strategy as the "out of the box way it works" that's closest to how Neo4j deals with TXs.  I also think the log compaction need is legitimate though, and instead of simple config to change the keying strategy, we'd need to configure the source to use a second "producer strategy" called, oh, something like nodesAndRels or something similar.   This producer strategy would produce many kafka messages (in one kafka TX) per neo4j TX.

This is probably not so simple to fix though as the original PR.

conker84:
@Solom00n & @moxious we already support transactions for the source, you just need to configure the `kafka.transaction.id` property and we leverage the Kafka transactions.

https://github.com/neo4j-contrib/neo4j-streams/blob/4442020941e4e8cb17703daab22966cdf8cdc74c/producer/src/main/kotlin/streams/kafka/KafkaEventRouter.kt#L95-L122

And it's used for batch of entities involved in a Neo4j transaction (i.e the `CREATE` statement that @moxious mentioned):

https://github.com/neo4j-contrib/neo4j-streams/blob/4442020941e4e8cb17703daab22966cdf8cdc74c/producer/src/main/kotlin/streams/StreamsTransactionEventHandler.kt#L274

So what we need at this point is the is `Source strategy` as David mentioned but the implementation should not be so problematic considering that we already support Kafka TXs

Wdyt?

moxious:
I see a lot of value in @Solom00n 's suggestions and think this would be a good thing to do, if you don't feel it's too large an amount of work right now given other priorities.

This has to be noted in the documentation though that kafka.transactional.id is required, and the plugin should fail to initialize with an appropriate error message if the user tells it to use this production strategy and doesn't give a transactional ID, because I think this feature *without* Kafka transactions is potentially a big problem.  With transactions I don't have the same concerns.

jexp:
I don't think we should use the internal id's here but better label + business-ids. (from constraints9
Then you can compact the streams even if they are coming from different databases.
Or a restored/comnpacted version of this database.

conker84:
I'm closing this we'll get back into it ASAP as it's a very important topic.

PR #317, opened 2020-05-21
Fixes neo4j-contrib/neo4j-streams#313 Add Timestamp, Header and Key

### LINKED ISSUE #313: Add support for reading Key, Headers and Timestamp of a message
## Feature description
Add the ability to Neo4J Streams (both Kafka Connect and Kafka Plugin) to read the headers, key and timestamp of a Kafka message rather than just only its value.

## Considered alternatives
Taking a look into the code, right now there is no way to access that information, so there is no possible alternative.

Giving Neo4J Streams the ability to read Headers, Key and Timestamp could add a lot of value, as not all the important information relies on the value of a Kafka message. Here you have some examples:

* **CDC Delete events:** Usually, when Kafka is used as a CDC platform, records deletion do have usually this format: `{headers:[], key:[], value: null}`, setting this message as a Kafka Tombstone record. That means that the record with the given key is now deleted. If Neo4J Streams only rely on the message value, there is no way to acknowledge the deletion of an event, as all the information it has is a null value. Adding the ability to read each message key Neo4J Streams could delete information from Neo4J when this messaging pattern is present.
* **Not every information is on the value:** The actual implementations of the sinks implementation assume that all the valuable information of each message relies on the message's value, and it doesn't have to be on that way. As keys in Kafka are used also for better partitioning strategies, some information could be stored just in the key rather than duplicating information both in the key and the value.
* **Access to message metadata:** Kafka messages can have some headers attached to them, and these fields are usually treated as metadata or control information that could be useful.
* **Timestamp could avoid out-of-order messages:** As stated in the documentation, Sink connectors can be set as parallel or not, depending on the importance of events ordering. That approach assumes that every consumed message happens before the next one, and that doesn't have to be that way. Giving the ability to access the record timestamp could give users information about whether or not update a Node or Relationship based on its timestamps.

Because of this, I propose the following:
* Make the header accessible from Cypher, adding `UNWIND {header} as header`.
* Make the key accessible from Cypher, adding `UNWIND {key} as key`.
* Make the record timestamp accessible from Cypher, adding `UNWIND {timestamp} as timestamp`.

Dionakra:
Fixes #313 

Add support to handle data from a message's timestamp, headers and key both in Neo4J Kafka Plugin and Kafka Connect Neo4J

Right now both Kafka Connect Neo4J and Neo4J Consumer Kafka Plugin only takes into account a message's value, but there is lots of useful information on the other Kafka message's parts as stated in the related issue.

Because of that, this pull request adds the message's timestamp, headers and key to the `event` object, so all that information is accessible from whoever needs it.

To do so, three fields have been added to the `event` object:
* **\_\_timestamp\_\_:** Message's timestamp (Long).
* **\_\_headers_\_:** Message's headers as a `Map<String, String>`. The type of the value can be anything as a Kafka Header is a `Map<String, Array<Byte>>`, but the value is usually a String or some sort of data with an easy String representation like booleans, integers, longs, doubles and so on. If a message's header is `null`, the `__headers__` field will be set as an empty object to avoid null references.
* **\_\_key_\_:** Message's key. It can be accessed as `event.__key__` if the key is a simple object, like a String, or as `event.__key__.field` if the key has some fields after extracting its schema with the deserializer. If the key is `null`, the `__key__` field will be set as an empty object to avoid null references.

As these changes could cause a naming collision between pre existing field names inside the `event` object, this change also provides a way to customize how these new fields are named.
* Neo4J Kafka Plugin
  * `streams.event.timestamp`,
  * `streams.event.headers`,
  * `streams.event.key`,
* Kafka Connect Neo4J
  * `neo4j.event.timestamp`
  * `neo4j.event.headers`
  * `neo4j.event.key`


conker84:
Hi @Dionakra thanks a lot for the effort that you put on this, I think that we should not merge it, there are multiple ways to achieve, outside the Streams Plugin, what you're trying to put here.
See Single Message Transformations in Kafka Connect or KSQL, that can allow the developer to pre-process the stream obtaining the same result.

Dionakra:
Hi @conker84 , 

Thank you, I didn't think about SMTs, you are completely right! Sorry for making you waste your time!

Anyway, that only applies to the Kafka Connect Sink, not the Neo4J Kafka plugin. If I wanted to use the Neo4J Kafka Plugin, what would be the recommended approach to access a message's headers and key?

Thank you.

conker84:
You can use the Kafka Connect framework as middleware, that manipulates your stream and prepare it with the necessary information that you need for the ingestion into Neo4j with te Streams Plugin.
This approach works for both Streams and Neo4j Kafka Connect Sink.

PR #320, opened 2020-06-03
add the sink lifecycle procedures

conker84:
This PR adds the sink lifecycle procedures allowing the user to start/stop/restart the sink without the need to physically restart the Neo4j instance.

**N.b.** Any changes made on the `neo4j.conf` file will not affect Streams conf.

- added `streams.sink.stop`
- added `streams.sink.start`
- added `streams.sink.restart`
- added tests

### DIFF
--- a/consumer/src/main/kotlin/streams/StreamsEventSinkExtensionFactory.kt
+++ b/consumer/src/main/kotlin/streams/StreamsEventSinkExtensionFactory.kt
@@ -72,6 +72,7 @@ class StreamsEventSinkExtensionFactory : KernelExtensionFactory<StreamsEventSink
                             StreamsSinkProcedures.registerStreamsSinkConfiguration(streamsSinkConfiguration)
                             StreamsSinkProcedures.registerStreamsEventConsumerFactory(eventSink.getEventConsumerFactory())
                             StreamsSinkProcedures.registerStreamsEventSinkConfigMapper(eventSink.getEventSinkConfigMapper())
+                            StreamsSinkProcedures.registerStreamsEventSink(eventSink)
                         } catch (e: Exception) {
                             streamsLog.error("Error initializing the streaming sink:", e)
                         }
--- a/consumer/src/main/kotlin/streams/StreamsEventSinkExtensionFactory.kt
+++ b/consumer/src/main/kotlin/streams/StreamsEventSinkExtensionFactory.kt
@@ -88,7 +88,7 @@ class StreamsEventSinkExtensionFactory : KernelExtensionFactory<StreamsEventSink
                 GlobalScope.launch(Dispatchers.IO) {
                     val success = StreamsUtils.blockUntilTrueOrTimeout(streamsSinkConfiguration.checkApocTimeout, streamsSinkConfiguration.checkApocInterval) {
                         val hasApoc = Neo4jUtils.hasApoc(db)
-                        if (streamsLog.isDebugEnabled) {
+                        if (!hasApoc && streamsLog.isDebugEnabled) {
                             streamsLog.debug("APOC not loaded yet, next check in ${streamsSinkConfiguration.checkApocInterval} ms")
                         }
                         hasApoc
--- a/consumer/src/main/kotlin/streams/StreamsEventSinkExtensionFactory.kt
+++ b/consumer/src/main/kotlin/streams/StreamsEventSinkExtensionFactory.kt
@@ -1,5 +1,8 @@
 package streams
 
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.GlobalScope
+import kotlinx.coroutines.launch
 import org.neo4j.kernel.availability.AvailabilityGuard
 import org.neo4j.kernel.availability.AvailabilityListener
 import org.neo4j.kernel.configuration.Config
@@ -81,6 +84,28 @@ class StreamsEventSinkExtensionFactory : KernelExtensionFactory<StreamsEventSink
         }
 
         private fun initSinkModule(streamsTopicService: StreamsTopicService, streamsSinkConfiguration: StreamsSinkConfiguration) {
+            if (streamsSinkConfiguration.checkApocTimeout > -1) {
+                GlobalScope.launch(Dispatchers.IO) {
+                    val success = StreamsUtils.blockUntilTrueOrTimeout(streamsSinkConfiguration.checkApocTimeout, streamsSinkConfiguration.checkApocInterval) {
+                        val hasApoc = Neo4jUtils.hasApoc(db)
+                        if (streamsLog.isDebugEnabled) {
+                            streamsLog.debug("APOC not loaded yet, next check in ${streamsSinkConfiguration.checkApocInterval} ms")
+                        }
+                        hasApoc
+                    }
+                    if (success) {
+                        initSink(streamsTopicService, streamsSinkConfiguration)
+                    } else {
+                        streamsLog.info("Streams Sink plugin not loaded as APOC are not installed")
+                    }
+                }
+            } else {
+                initSink(streamsTopicService, streamsSinkConfiguration)
+            }
+
+        }
+
+        private fun initSink(streamsTopicService: StreamsTopicService, streamsSinkConfiguration: StreamsSinkConfiguration) {
             streamsTopicService.clearAll()
             streamsTopicService.setAll(streamsSinkConfiguration.topics)
             eventSink.start()
--- a/consumer/src/main/kotlin/streams/procedures/StreamsSinkProcedures.kt
+++ b/consumer/src/main/kotlin/streams/procedures/StreamsSinkProcedures.kt
@@ -1,7 +1,8 @@
 package streams.procedures
 
+import org.apache.commons.lang3.exception.ExceptionUtils
 import org.neo4j.graphdb.GraphDatabaseService
-import org.neo4j.kernel.configuration.Config
+import org.neo4j.kernel.internal.GraphDatabaseAPI
 import org.neo4j.logging.Log
 import org.neo4j.procedure.Context
 import org.neo4j.procedure.Description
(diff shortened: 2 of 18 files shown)


consumer/src/main/kotlin/streams/procedures/StreamsSinkProcedures.kt
```diff
@@ -27,13 +39,31 @@ class StreamsSinkProcedures {
...
         val properties = config?.mapValues { it.value.toString() } ?: emptyMap()
-        val configuration = StreamsSinkProcedures.streamsEventSinkConfigMapper.convert(config = properties)
+        val configuration = streamsEventSinkConfigMapper.convert(config = properties)
 
         val data = readData(topic, config ?: emptyMap(), configuration)
 
         return data.map { StreamResult(mapOf("data" to it)) }.stream()
     }
 
+    @Procedure("streams.sink.start")
```
moxious, line 49: This should return a boolean of whether or not the operation worked, rather than returning nothing.

moxious: Additionally, what will happen if this is called when no streams sink configuration is specified?  (Exception should be the result)

consumer/src/main/kotlin/streams/procedures/StreamsSinkProcedures.kt
```diff
@@ -27,13 +39,31 @@ class StreamsSinkProcedures {
...
+    fun sinkStart(): Unit {
+        checkEnabled()
+        streamsEventSink?.start()
+    }
+
+    @Procedure("streams.sink.stop")
```
moxious, line 55: This should return a boolean to indicate success/failure.   You can return true ("yes, I stopped it") if it never began, or wasn't configured.  Because the user asked to stop, and that succeeded in those cases.

consumer/src/main/kotlin/streams/procedures/StreamsSinkProcedures.kt
```diff
@@ -27,13 +39,31 @@ class StreamsSinkProcedures {
...
+    fun sinkStop(): Unit {
+        checkEnabled()
+        streamsEventSink?.stop()
+    }
+
+    @Procedure("streams.sink.restart")
```
moxious, line 61: return the start operation result

moxious:
Request from the account team to add one last "status" proc which will return the status.   I'd recommend this maybe return a map of the SINK ONLY configuration, along with the boolean true/false is it running.   This will allow users to inspect the effective config, and know whether it's active.

conker84:
> Request from the account team to add one last "status" proc which will return the status. I'd recommend this maybe return a map of the SINK ONLY configuration, along with the boolean true/false is it running. This will allow users to inspect the effective config, and know whether it's active.
> 
> This last request is _optional_ on this PR and can be handled separately -- because it's not realted to unblocking the key issue here, but I think it's a good ask if it's not a heavy lift

Is it ok this kind of info?

```
+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| name                        | value                                                                                                                                                 |
+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| "enabled"                   | true                                                                                                                                                  |
| "procedures.enabled"        | true                                                                                                                                                  |
| "sink.polling.interval"     | 10000                                                                                                                                                 |
| "topics"                    | {relPatternTopics -> {}, cudTopics -> [], invalid -> [], cdcSourceIdTopics -> [], cypherTopics -> {}, cdcSchemaTopics -> [], nodePatternTopics -> {}} |
| "error.config"              | {}                                                                                                                                                    |
| "source.id.strategy.config" | {idName -> "sourceId", labelName -> "SourceEvent"}                                                                                                    |
+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
6 rows
```

moxious:
This does do what I had in mind.  I've asked the account team to comment directly on this as well though before we finalize.

rhaces:
> Is it ok this kind of info?

For the immediate need, the goal would be to see if someone sent `streams.sink.stop`, so if `"enabled":false` means that, then I'm good.
What I would love as an ultimate solution would be to be able to see all the topics the sink is configured to be subscribed and if it is currently subscribed or not... something like:
```
{
  "enabled": true,
  "topics": [
    {"topic a": "subscribed"},
    {"topic b": "unsubscribed"},
    {"topic c": "subscribed"},
    {"topic d": "unsubscribed"}
  ]
}
```

consumer/src/main/kotlin/streams/procedures/StreamsSinkProcedures.kt
conker84: @moxious I'll add the exception in case is not a LEADER or a single instance configuration.
About the config, we'll reuse the same configuration provided at startup time, so no change can be provided, we can work on a new PR where the user can provide the new config via the procedure wdyt?

moxious: All good -- except defer the "provide new config via procedure".  This is a big complicated point for all plugins, how to support dynamic config.  It'll cause separate problems if the config at runtime doesnt' match what's in the config file.  I don't want to get into that with the functionality here until we have a better solution for dynamic config of plugins, which like we discussed before isn't really a neo4j-streams issue.    If we try to do dynamic config without the conf file, it's going to get really complicated quickly with feature requests & bugs, and we need a better overall approach for that rather than adding as an incremental feature.

conker84:
Hi @rhaces we can provide this info:

```
+------------------------------------------------------------------------------------------------+
| name                                      | value                                              |
+------------------------------------------------------------------------------------------------+
| "streams.sink.errors"                     | {}                                                 |
| "streams.sink.source.id.strategy.config"  | {idName -> "sourceId", labelName -> "SourceEvent"} |
| "streams.sink.topic.cypher"               | {}                                                 |
| "streams.sink.topic.cud"                  | []                                                 |
| "streams.sink.topic.cdc.schema"           | []                                                 |
| "streams.sink.topic.cdc.sourceId"         | []                                                 |
| "streams.sink.topic.pattern.node"         | {}                                                 |
| "streams.sink.topic.pattern.relationship" | {}                                                 |
| "invalid_topics"                          | []                                                 |
+------------------------------------------------------------------------------------------------+
```

Following a description of each field:

 | Config Name | Description | 
 | ------------- | ------------- |
 | invalid_topics | return a list of invalid topics | 
 | streams.sink.topic.pattern.relationship | return a Map<K,V> where the K is the topic name and V is the provided pattern | 
 | streams.sink.topic.cud | return a list of topics defined for the CUD format | 
 | streams.sink.topic.cdc.sourceId | return a list of topics defined for the CDC SourceId strategy | 
 | streams.sink.topic.cypher | return a Map<K,V> where the K is the topic name and V is the provided Cypher Query | 
 | streams.sink.topic.cdc.schema | return a list of topics defined for the CDC Schema strategy | 
 | streams.sink.topic.pattern.node | return a Map<K,V> where the K is the topic name and V is the provided pattern | 
 | streams.sink.errors | return a Map<K,V> where the K sub property name, and V is the value | 
 | streams.sink.source.id.strategy.config | returns the config for the SourceId CDC strategy |

Moreover, I added another procedure `streams.sink.status` that returns the status of the Sink (`RUNNING/STOPPED`)

consumer/src/main/kotlin/streams/procedures/StreamsSinkProcedures.kt
conker84: done

consumer/src/main/kotlin/streams/procedures/StreamsSinkProcedures.kt
conker84: done, but if there is an error in the stop I'll return it with the error

conker84:
@moxious I added the docs and please look at the comment above to check the changes since your feedback

consumer/src/main/kotlin/streams/procedures/StreamsSinkProcedures.kt
```diff
@@ -27,13 +44,86 @@ class StreamsSinkProcedures {
...
+        return (configMap + topicMap + invalidTopics)
+                .entries.stream()
+                .map { StreamsSinkDTO(it.key, it.value) }
+    }
+
+    @Procedure("streams.sink.status")
+    fun sinkStatus(): Stream<StreamsSinkDTO> {
+        checkEnabled()
+        checkLeader()
+        return Stream.of(StreamsSinkDTO("status", streamsEventSink?.status() ?: StreamsPluginStatus.UNKNOWN))
```
moxious, line 118: Failed local testing for me with: 

```
Cannot convert streams.events.StreamsPluginStatus to AnyValue
```

moxious: When I did a stop and then a start, I got the same thing (but stop worked and returned true)

conker84: Yes, it odd because the two test passes:

https://github.com/neo4j-contrib/neo4j-streams/blob/e8dbd7982c47e4178b59a52f48b18220016db602/consumer/src/test/kotlin/integrations/kafka/KafkaStreamsSinkProcedures.kt#L276-L311

Let me covert the status to string

conker84:
@moxious pr updated

consumer/src/main/kotlin/streams/StreamsEventSinkExtensionFactory.kt
```diff
@@ -81,6 +84,28 @@ class StreamsEventSinkExtensionFactory : KernelExtensionFactory<StreamsEventSink
         }
 
         private fun initSinkModule(streamsTopicService: StreamsTopicService, streamsSinkConfiguration: StreamsSinkConfiguration) {
+            if (streamsSinkConfiguration.checkApocTimeout > -1) {
+                GlobalScope.launch(Dispatchers.IO) {
+                    val success = StreamsUtils.blockUntilTrueOrTimeout(streamsSinkConfiguration.checkApocTimeout, streamsSinkConfiguration.checkApocInterval) {
+                        val hasApoc = Neo4jUtils.hasApoc(db)
+                        if (streamsLog.isDebugEnabled) {
```
moxious, line 91: only if hasApoc is false.   Otherwise this will output the message every time, even after it succeeded.

conker84: done, thanks for spotting this!

PR #328, opened 2020-07-15
add sink lifecyle procedures

conker84:
Fixes #320 in branch 4.0

Provide the same set of feature as we did in branch `3.5`

- added `streams.sink.stop`
- added `streams.sink.start`
- added `streams.sink.restart`
- added tests

### DIFF
--- a/common/src/main/kotlin/streams/utils/Neo4jUtils.kt
+++ b/common/src/main/kotlin/streams/utils/Neo4jUtils.kt
@@ -37,6 +37,21 @@ object Neo4jUtils {
         }
     }
 
+    fun hasApoc(db: GraphDatabaseAPI): Boolean {
+        try {
+            return db.execute("RETURN apoc.version() AS version") {
+                it.columnAs<String>("version").next()
+                true
+            }
+        } catch (e: QueryExecutionException) {
+            if (e.statusCode.equals("Neo.ClientError.Statement.SyntaxError", ignoreCase = true)
+                    && e.message!!.contains("Unknown function", ignoreCase = true)) {
+                return false
+            }
+            throw e
+        }
+    }
+
     fun getLogService(db: GraphDatabaseAPI): LogService {
         return db.dependencyResolver
                 .resolveDependency(LogService::class.java)
--- a/common/src/main/kotlin/streams/utils/Neo4jUtils.kt
+++ b/common/src/main/kotlin/streams/utils/Neo4jUtils.kt
@@ -37,19 +37,13 @@ object Neo4jUtils {
         }
     }
 
-    fun hasApoc(db: GraphDatabaseAPI): Boolean {
-        try {
-            return db.execute("RETURN apoc.version() AS version") {
-                it.columnAs<String>("version").next()
-                true
-            }
-        } catch (e: QueryExecutionException) {
-            if (e.statusCode.equals("Neo.ClientError.Statement.SyntaxError", ignoreCase = true)
-                    && e.message!!.contains("Unknown function", ignoreCase = true)) {
-                return false
-            }
-            throw e
+    fun hasApoc(db: GraphDatabaseAPI): Boolean = try {
+         db.execute("RETURN apoc.version() AS version") {
+            it.columnAs<String>("version").next()
+            true
         }
+    } catch (e: QueryExecutionException) {
+        false
     }
 
     fun getLogService(db: GraphDatabaseAPI): LogService {
--- a/consumer/src/main/kotlin/streams/kafka/KafkaEventSink.kt
+++ b/consumer/src/main/kotlin/streams/kafka/KafkaEventSink.kt
@@ -20,6 +20,7 @@ import streams.StreamsEventSinkQueryExecution
 import streams.StreamsSinkConfiguration
 import streams.StreamsTopicService
 import streams.config.StreamsConfig
+import streams.events.StreamsPluginStatus
 import streams.service.errors.ErrorService
 import streams.service.errors.KafkaErrorService
 import streams.utils.KafkaValidationUtils.getInvalidTopicsError
@@ -77,6 +78,12 @@ class KafkaEventSink(private val config: StreamsConfig,
     }
 
     override fun start() { // TODO move to the abstract class
+        if (StreamsPluginStatus.RUNNING == status()) {
+            if (log.isDebugEnabled) {
+                log.debug("Kafka Sink is already started.")
+            }
+            return
+        }
         val streamsConfig = StreamsSinkConfiguration.from(config, db.databaseName())
         val topics = streamsTopicService.getTopics()
         val isWriteableInstance = Neo4jUtils.isWriteableInstance(db)
@@ -137,7 +144,7 @@ class KafkaEventSink(private val config: StreamsConfig,
                         }
                         TimeUnit.SECONDS.toMillis(1)
                     } else {
-                        val timeMillis = TimeUnit.MINUTES.toMillis(5)
+                        val timeMillis = TimeUnit.MINUTES.toMillis(3)
                         if (log.isDebugEnabled) {
                             log.debug("Not in a writeable instance, new check in $timeMillis millis")
                         }
@@ -165,6 +172,11 @@ class KafkaEventSink(private val config: StreamsConfig,
                 log.warn(getInvalidTopicsError(eventConsumer.invalidTopics()))
             }
         }, UninitializedPropertyAccessException::class.java)
+    }
+    
+    override fun status(): StreamsPluginStatus = when (this::job.isInitialized && this.job.isActive) {
+        true -> StreamsPluginStatus.RUNNING
+        else -> StreamsPluginStatus.STOPPED
     }
 
 }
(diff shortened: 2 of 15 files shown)


jexp:
Are these meant to be per database or global?

common/src/main/kotlin/streams/utils/Neo4jUtils.kt
```diff
@@ -37,6 +37,21 @@ object Neo4jUtils {
...
     }
 
+    fun hasApoc(db: GraphDatabaseAPI): Boolean {
+        try {
+            return db.execute("RETURN apoc.version() AS version") {
+                it.columnAs<String>("version").next()
+                true
+            }
+        } catch (e: QueryExecutionException) {
+            if (e.statusCode.equals("Neo.ClientError.Statement.SyntaxError", ignoreCase = true)
```
jexp, line 47: too specific
I would return false in any error case

consumer/src/main/kotlin/streams/kafka/KafkaEventSink.kt
```diff
@@ -137,7 +144,7 @@ class KafkaEventSink(private val config: StreamsConfig,
                         }
                         TimeUnit.SECONDS.toMillis(1)
                     } else {
-                        val timeMillis = TimeUnit.MINUTES.toMillis(5)
```
jexp, line 140: 5 or 3 minutes both sound quite long, or?

jexp: is that configureable?

consumer/src/main/kotlin/streams/procedures/StreamsSinkProcedures.kt
```diff
@@ -1,21 +1,34 @@
...
 import streams.config.StreamsConfig
+import streams.events.StreamsPluginStatus
+import streams.extensions.toPointCase
+import streams.serialization.JSONUtils
+import streams.utils.Neo4jUtils
+import java.util.stream.Collectors
 import java.util.stream.Stream
 
 class StreamResult(@JvmField val event: Map<String, *>)
+class StreamsSinkDTO(@JvmField val name: String, @JvmField val value: Any?)
```
jexp, line 22: that's a really odd name as it leaks implementation details into the API

consumer/src/main/kotlin/streams/procedures/StreamsSinkProcedures.kt
```diff
@@ -35,6 +48,80 @@ class StreamsSinkProcedures {
...
+                    Stream.of(StreamsSinkDTO("exception", ExceptionUtils.getStackTrace(e))))
+        }
+    }
+
+    @Procedure("streams.sink.stop")
+    fun sinkStop(): Stream<StreamsSinkDTO> {
+        checkEnabled()
+        checkLeader()
+        try {
+            streamsEventSink?.stop()
```
jexp, line 70: what happens with the status that is managed by the kernel-extension-factory?

Will it try to stop / start the sink again if it's already started or stopped? Can we make sure that still works.

.e.g. someone calls sink.stop and then the database triggers a stop, will that be handled gracefully

or a restart of the db in the cluster

consumer/src/main/kotlin/streams/procedures/StreamsSinkProcedures.kt
```diff
@@ -35,6 +48,80 @@ class StreamsSinkProcedures {
...
+            return stopped.stream()
+        }
+        return sinkStart()
+    }
+
+    @Procedure("streams.sink.config")
+    fun sinkConfig(): Stream<StreamsSinkDTO> {
+        checkEnabled()
+        checkLeader()
+        val configMap = JSONUtils.asMap(streamsSinkConfiguration)
```
jexp, line 93: are there any passwords/credentials that we need to filter out/mask?

consumer/src/main/kotlin/streams/procedures/StreamsSinkProcedures.kt
```diff
@@ -35,6 +48,80 @@ class StreamsSinkProcedures {
...
+                .filterKeys { it != "topics" && it != "enabled" && it != "proceduresEnabled" && !it.startsWith("check") }
+                .mapKeys { it.key.toPointCase() }
+                .mapKeys {
+                    when (it.key) {
+                        "error.config" -> "streams.sink.errors"
```
jexp, line 98: looks as if this config listing / renaming (parse / output) should be in a central place and not hidden away in the procedure

so if something changes there we need to change it only in one place.

consumer/src/main/kotlin/streams/procedures/StreamsSinkProcedures.kt
```diff
@@ -35,6 +48,80 @@ class StreamsSinkProcedures {
...
+                .mapKeys { it.key.key }
+        val invalidTopics = mapOf("invalid_topics" to streamsSinkConfiguration.topics.invalid)
+        return (configMap + topicMap + invalidTopics)
+                .entries.stream()
+                .map { StreamsSinkDTO(it.key, it.value) }
+    }
+
+    @Procedure("streams.sink.status")
+    fun sinkStatus(): Stream<StreamsSinkDTO> {
+        checkEnabled()
```
jexp, line 113: do these functions have any return values?

or will they just error if something goes wrong?

consumer/src/main/kotlin/streams/procedures/StreamsSinkProcedures.kt
```diff
@@ -35,6 +48,80 @@ class StreamsSinkProcedures {
...
+        checkLeader()
+        val value = (streamsEventSink?.status() ?: StreamsPluginStatus.UNKNOWN).toString()
+        return Stream.of(StreamsSinkDTO("status", value))
+    }
+
+    private fun checkLeader() {
+        if (!Neo4jUtils.isWriteableInstance(db as GraphDatabaseAPI)) {
```
jexp, line 120: should that be returned as an exception or as part of the status "error" -> "You can use it only in the LEADER or in a single instance configuration."

also what is `it`.
and this is another piece of code that should probably live in a more central place?

doc/asciidoc/procedures/index.adoc
```diff
@@ -241,3 +241,64 @@ Supported deserializers are: `org.apache.kafka.common.serialization.ByteArrayDes
 
 |===
 
+=== Streams Sink Lifecycle procedure
```
jexp, line 244: can we add a usage example of an (textual) output of such a lifecycle?

```
listconfig
stop
status
start
status
restart
status on non leader
```

jexp: we can also generate that from a test an just adoc-include the file.

jexp (approved):
please have a look at my comments, looks good otherwise.

common/src/main/kotlin/streams/utils/Neo4jUtils.kt
conker84: done

consumer/src/main/kotlin/streams/kafka/KafkaEventSink.kt
conker84: I'll add the configuration in the next PR where I'll add all the changes made in #344

consumer/src/main/kotlin/streams/procedures/StreamsSinkProcedures.kt
conker84: changed

consumer/src/main/kotlin/streams/procedures/StreamsSinkProcedures.kt
conker84: Great question! I'll manage this in the next PR where I'll add all the changes made in #344

consumer/src/main/kotlin/streams/procedures/StreamsSinkProcedures.kt
conker84: no, here we have only properties regarding `streams.*`

consumer/src/main/kotlin/streams/procedures/StreamsSinkProcedures.kt
conker84: moved into a method inside the `StreamsSinkConfiguration` class

consumer/src/main/kotlin/streams/procedures/StreamsSinkProcedures.kt
conker84: they just throw an exception if something goes wrong

consumer/src/main/kotlin/streams/procedures/StreamsSinkProcedures.kt
conker84: I changed it as you requested, but I left it here as is something specific for procedures

doc/asciidoc/procedures/index.adoc
conker84: done

conker84:
> Are these meant to be per database or global?

per database

conker84:
@jexp I updated the PR with your feedback. Thank you!

moxious:
OK to merge?

consumer/src/main/kotlin/streams/procedures/StreamsSinkProcedures.kt
```diff
@@ -35,6 +50,71 @@ class StreamsSinkProcedures {
...
+                log?.error("Cannot stopped the Sink because of the following exception", e)
+                Stream.concat(sinkStatus(),
+                        Stream.of(KeyValueResult("exception", ExceptionUtils.getStackTrace(e))))
+            }
+        }
+    }
+
+    @Procedure("streams.sink.restart")
+    fun sinkRestart(): Stream<KeyValueResult> {
+        val stopped = sinkStop().collect(Collectors.toList())
```
jexp, line 85: this could be simplified but is not important

jexp (approved):

PR #331, opened 2020-07-17
Migrated to Antora documentation

mroiter-larus:
In order to locally preview the Neo4j Spark connector documentation built with Antora do the following steps:

1. open a terminal window and be sure to be at the root of the project
2. **This step has to be executed only the first time**: `cd doc/docs && npm install express`. 
3. run `cd doc` (if you have not run the command at step 2) or run `cd ..` (if you have run step 2)
4. run `antora docs.yml`
5. run `npm start`
6. browse to localhost:8000


conker84:
How can we integrate this into our release pipeline? It's something that we can delegate to maven?

moxious:
@mroiter-larus @conker84 a thing to consider - the build pipeline which publishes docs live to the site appears to require a directory called `in` at the end of the build step.   The previous build step does `mvn clean packageHTML` (so it appears configured by @mneedham ) -- so it would be desirable to make that maven goal work (or some similar maven goal we can configure) to keep the deploy pipeline similar.

mroiter-larus:
@moxious i would like to use a maven plugin in order to run `npm install` automatically when the `mvn clean package` or `mvn clean install` is executed. This should install all dependencies Antora needs. But in order to generate the Antora documentation site another step is necessary: `antora <playbook-name>.yml`.
I think this is not executable via Maven.
However, are you sure that the previous build step does `mvn clean packageHTML`? From what i know `packageHTML` is a Gradle task.

moxious:
@mroiter-larus here are a picture of the build steps.   Yes to packageHTML - this is how it's currently working.   We *can* change this though, it just requires more configuration in the automatic build

Screen Shot 2020-07-22 at 2 02 47 PM

moxious:
I should point out that we can still get most of the way and merge the PR -- it just has to be possible to generate documentation with some set of command line steps, and then we can rig the build chain to produce that.  In the end all that's needed is a set of static html assets to upload to a bucket

mroiter-larus:
@moxious PR updated! 
Now when you run `mvn clean install` Maven will execute also `npm install` so Antora and all the necessary dependencies will be installed automatically.
That said, i think the only build step that has to be changed is the last one. Now it should do `npm run build:docs` or equivalently `antora docs.yml`. In the same way as the previous Gradle command, the new one has to be executed from the `doc` folder.

moxious:
LGTM.   Merge when ready, then maybe we can work out the build system bits and publish new documentation

PR #333, opened 2020-07-28
Added support for secured URI schemes

### LINKED ISSUE #325: Neo4j 4 for Aura - Integration testing
This ticket is not a bug - just a task to do some compatibility / integration testing with Neo4j 4 running on Aura, which is coming available soon.

Specific changes that I know about in Aura 4 -- which may impact the connect worker and that need testing:

- The `neo4j` user no longer has the admin role
- `CALL dbms.listConfig()` and some other stored procedures as a result cannot be called from the client side
- If we have any versioning switches, these need careful review because `CALL dbms.components()` returns custom information.

mroiter-larus:
Fixes #325 

Added support for Neo4j secured URI schemes

### DIFF
--- a/kafka-connect-neo4j/src/main/kotlin/streams/kafka/connect/sink/Neo4jService.kt
+++ b/kafka-connect-neo4j/src/main/kotlin/streams/kafka/connect/sink/Neo4jService.kt
@@ -42,20 +42,24 @@ class Neo4jService(private val config: Neo4jSinkConnectorConfig):
 
     init {
         val configBuilder = Config.builder()
-        if (this.config.encryptionEnabled) {
-            configBuilder.withEncryption()
-            val trustStrategy: Config.TrustStrategy = when (this.config.encryptionTrustStrategy) {
-                Config.TrustStrategy.Strategy.TRUST_ALL_CERTIFICATES -> Config.TrustStrategy.trustAllCertificates()
-                Config.TrustStrategy.Strategy.TRUST_SYSTEM_CA_SIGNED_CERTIFICATES -> Config.TrustStrategy.trustSystemCertificates()
-                Config.TrustStrategy.Strategy.TRUST_CUSTOM_CA_SIGNED_CERTIFICATES -> Config.TrustStrategy.trustCustomCertificateSignedBy(this.config.encryptionCACertificateFile)
-                else -> {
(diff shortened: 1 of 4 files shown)


kafka-connect-neo4j/src/main/kotlin/streams/kafka/connect/sink/Neo4jService.kt
```diff
@@ -42,20 +42,24 @@ class Neo4jService(private val config: Neo4jSinkConnectorConfig):
...
-        if (this.config.encryptionEnabled) {
-            configBuilder.withEncryption()
-            val trustStrategy: Config.TrustStrategy = when (this.config.encryptionTrustStrategy) {
-                Config.TrustStrategy.Strategy.TRUST_ALL_CERTIFICATES -> Config.TrustStrategy.trustAllCertificates()
-                Config.TrustStrategy.Strategy.TRUST_SYSTEM_CA_SIGNED_CERTIFICATES -> Config.TrustStrategy.trustSystemCertificates()
-                Config.TrustStrategy.Strategy.TRUST_CUSTOM_CA_SIGNED_CERTIFICATES -> Config.TrustStrategy.trustCustomCertificateSignedBy(this.config.encryptionCACertificateFile)
-                else -> {
-                    throw ConfigException(Neo4jSinkConnectorConfig.ENCRYPTION_TRUST_STRATEGY, this.config.encryptionTrustStrategy.toString(), "Encryption Trust Strategy is not supported.")
+
+        if (!this.config.hasSecuredURI()) {
```
moxious, line 46: I think this code is good and fine -- but setting explicit trust settings when using URLs like neo4j+s and neo4j+ssc is also not strictly needed.

Our drivers team has said that if you use neo4j+s or neo4j+ssc, then these settings are ignored.  If they conflict you could get an error (like for example using neo4j+s and also TRUST_ALL_CERTIFICATES, which doesnt' make sense) -- but if they don't conflict, it's not really needed.   Still this is OK

conker84:
@mroiter-larus can you please add a link into the Connect docs about supported uri schemes? So we can simplify the navigation for the user.

mroiter-larus:
@conker84 yes, sure!

conker84:
@mroiter-larus thank you

PR #353, opened 2020-09-11
Fixes to page titles in generated docs

adam-cowley:
I've changed the page title levels so the generated docs don't have Untitled in the page title.

Could you please double check that the content is still correct?

I've added a new npm command for previewing the docs - they will rebuild as you make changes so you can view them at localhost:8000

```
npm start
```


adam-cowley:
The streams.adoc file was causing warnings in the build process that will break future builds of the docs.  It doesn't seem to be referenced anywhere so I'm assuming it is OK to remove.  Could you please confirm?  

Here is the contents:

```adoc
include::producer.adoc[]

include::consumer.adoc[]

include::procedures.adoc[]
```

conker84 (approved):

doc/docs/modules/ROOT/pages/producer.adoc
```diff
@@ -1,38 +1,38 @@
+= Source: Neo4j -> Kafka
```
mroiter-larus, line 1: Does this title needs the prefix "Neo4j Streams -" like the other two chapters about Sink and Procedures modules?

adam-cowley: No it shouldn't need the prefix - having the prefix doesn't really add any value to the page

mroiter-larus (approved):

mroiter-larus (approved):

mroiter-larus (approved):

mroiter-larus:
@adam-cowley LGTM

PR #368, opened 2020-10-27
sink performances improvements

conker84:
This PR addresses the improvements that we discussed about the performance document.

### DIFF
--- a/consumer/src/main/kotlin/streams/kafka/KafkaEventSink.kt
+++ b/consumer/src/main/kotlin/streams/kafka/KafkaEventSink.kt
@@ -24,6 +24,7 @@ import streams.StreamsSinkConfiguration
 import streams.StreamsSinkConfigurationConstants
 import streams.StreamsTopicService
 import streams.events.StreamsPluginStatus
+import streams.serialization.JSONUtils
 import streams.utils.KafkaValidationUtils.getInvalidTopicsError
 import streams.utils.Neo4jUtils
 import streams.utils.StreamsUtils
@@ -62,12 +63,12 @@ class KafkaEventSink(private val config: Config,
 
     override fun getEventConsumerFactory(): StreamsEventConsumerFactory {
         return object: StreamsEventConsumerFactory() {
-            override fun createStreamsEventConsumer(config: Map<String, String>, log: Log): StreamsEventConsumer {
+            override fun createStreamsEventConsumer(config: Map<String, String>, log: Log, topics: Set<Any>): StreamsEventConsumer {
                 val kafkaConfig = KafkaSinkConfiguration.from(config)
                 return if (kafkaConfig.enableAutoCommit) {
-                    KafkaAutoCommitEventConsumer(kafkaConfig, log)
+                    KafkaAutoCommitEventConsumer(kafkaConfig, log, topics as Set<String>)
                 } else {
-                    KafkaManualCommitEventConsumer(kafkaConfig, log)
+                    KafkaManualCommitEventConsumer(kafkaConfig, log, topics as Set<String>)
                 }
             }
         }
@@ -100,8 +101,7 @@ class KafkaEventSink(private val config: Config,
             }
             log.info("Starting the Kafka Sink")
             eventConsumer = getEventConsumerFactory()
-                    .createStreamsEventConsumer(config.raw, log)
-                    .withTopics(streamsTopicService.getTopics())
+                    .createStreamsEventConsumer(config.raw, log, streamsTopicService.getTopics())
             job = createJob()
         }
         if (isWriteableInstance) {
@@ -136,7 +136,6 @@ class KafkaEventSink(private val config: Config,
 
     private fun createJob(): Job {
         log.info("Creating Sink daemon Job")
-        val hashcode = this.hashCode()
         return GlobalScope.launch(Dispatchers.IO) { // TODO improve exception management
             try {
                 eventConsumer.start()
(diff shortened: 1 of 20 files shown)


consumer/src/main/kotlin/streams/kafka/KafkaManualCommitEventConsumer.kt
```diff
@@ -3,15 +3,50 @@ package streams.kafka
...
+        }
+        super.stop()
+    }
 
     private fun commitData(commit: Boolean, topicMap: Map<TopicPartition, OffsetAndMetadata>) {
         if (commit) {
-            consumer.commitSync(topicMap)
+            if (asyncCommit) {
+                if (log.isDebugEnabled) {
+                    log.debug("Committing data in async")
```
moxious, line 34: might want to remove these debug log statements for the final one.  It should be clear which happens by configuration, and this method gets called for every read, so this could be quite some logspam outside of testing

consumer/src/main/kotlin/streams/kafka/KafkaEventSink.kt
```diff
@@ -148,7 +147,7 @@ class KafkaEventSink(private val config: Config,
                             }
                             queryExecution.writeForTopic(topic, data)
                         }
-                        TimeUnit.SECONDS.toMillis(1)
+                        streamsConfig.pollInterval
```
moxious, line 150: Below, you set a new default poll interval of 100ms, where it looks like the old one was 1ms.  Are you sure this is a better new default?  (I'm not sure myself, I just mean this introduces potential latency, but also avoids "busy waiting"). 

Wondering if you saw in the flamegraph that too much busy waiting was happening here and that's why this change, but I'm sensitive to changes in the default, because they can affect other workflows.  If you've got a specific reason then it's ok

consumer/src/main/kotlin/streams/kafka/KafkaManualCommitEventConsumer.kt
```diff
@@ -3,15 +3,50 @@ package streams.kafka
...
+                }
+                consumer.commitAsync(topicMap) { offsets: MutableMap<TopicPartition, OffsetAndMetadata>, exception: Exception? ->
+                    if (exception != null) {
+                        log.warn("""
+                            |These offsets `$offsets`
```
moxious, line 39: message should maybe mention topic name; user may have many topics configured, and offsets are just integers, right?

doc/asciidoc/consumer/configuration.adoc
```diff
@@ -23,11 +25,47 @@ streams.check.apoc.timeout=<ms to await for APOC being loaded, default -1 skip t
...
+to the caller).
+
+That means, the `commitSync` is a **blocking** method with an interal retry mechanism,
+that can affect the performance of the ingestion because a new batch of messages
+will be processed only when the commit ended.
+
+On the other hand `commitAsync` is an asynchronous call (so it will not block)
+and does not provide an internal retry mechanism.
+
+.Trade-offs: latency vs. data consistency
```
moxious, line 64: I want to make an argument that the default should be async commit, not sync.

1. Async is faster, because we don't have to wait for word back from Kafka
2. When an error occurs in sync commit, we can't really do much about it other than log it.
3. Most user complaints have come from throughput speed, not from commit semantics.  In fact, it's pretty usual for kafka clients to be "fire and forget".   There are of course exceptions.

The only situation I can think about where commitSync is better (after reading the docs) is when you're making a complex set of dependent changes.  For example in a CDC scenario, it may be very important that kafka got the first set of CDC changes before publishing the next set, if "order of application" matters.  If you commit async, I guess there is a risk that commit 2 could succeed before commit 1 and that could be bad.

But that's all in the direction of pushing data Neo4j -> Kafka.

In the case of consuming from kafka, is the following scenario possible:

* Batch 1 comes into Neo4j (finishing with offset 10 let's say).   CommitAsync for "A" starts.
* Batch 2 comes into Neo4j (finishing with offset 20 let's say).  CommitAsync for "B" starts.
* Commit for B finishes
* Commit for A finishes
* (Kafka now thinks our partition offset is 10, when it's really 20)
* Next poll pulls old data we've already seen

Is that possible?

consumer/src/main/kotlin/streams/kafka/KafkaEventSink.kt
conker84: No the 100ms is a random number for the tests, I defined the new one a 0 that should work better as default.

consumer/src/main/kotlin/streams/kafka/KafkaManualCommitEventConsumer.kt
conker84: `offsets` is a map of `<TopicPartition, OffsetAndMetadata>` so we print the information about:
- topic
- partition
- offset
- metadata

doc/asciidoc/consumer/configuration.adoc
conker84: From the official documentation seems that is not a problem:
https://docs.confluent.io/2.0.0/clients/consumer.html#asynchronous-commits
That use case in mi knowledge can happen only when a new consumer with the same `group.id` (ie. from a different threat) starts to consume the data, but it's a bad practice.

conker84: I used the default `false` only for backwards compatibility. WDTY?

moxious: I think I need to write more docs into the architecture section, specific to performance tuning.  It seems like keeping it sync is the best way to not suprise anyone, but now that you have the data to indicate that the performance is better with commitAsync, I think a lot of people will be interested in that.

conker84: Yes totally

moxious (approved):

moxious:
Ready to merge, but will wait a bit pending internal discussions with field engineers about way forward.

PR #369, opened 2020-10-30
sink performance improvements

conker84:
This PR addresses the improvements that we discussed about the performance document (for branch `4.0`).

### DIFF
--- a/consumer/src/main/kotlin/streams/kafka/KafkaManualCommitEventConsumer.kt
+++ b/consumer/src/main/kotlin/streams/kafka/KafkaManualCommitEventConsumer.kt
@@ -1,17 +1,87 @@
 package streams.kafka
 
+import org.apache.kafka.clients.consumer.CommitFailedException
+import org.apache.kafka.clients.consumer.ConsumerRebalanceListener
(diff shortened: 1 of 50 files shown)


consumer/src/main/kotlin/streams/kafka/KafkaManualCommitEventConsumer.kt
```diff
@@ -1,17 +1,87 @@
...
+        } catch (e: WakeupException) {
+            // we're shutting down, but finish the commit first and then
+            // rethrow the exception so that the main loop can exit
+            doCommitSync()
+            throw e
+        } catch (e: CommitFailedException) {
+            // the commit failed with an unrecoverable error. if there is any
+            // internal state which depended on the commit, you can clean it
+            // up here. otherwise it's reasonable to ignore the error and go on
+            log.debug("Commit failed", e)
```
moxious, line 45: log.warn?   It may not be critical but user should know

conker84: Done

conker84:
@moxious I'm merging this so we can prepare for the release

PR #372, opened 2020-11-18
Fixes #349: Provide option for streams.publish to operate synchronously

### LINKED ISSUE #349: Provide option for streams.publish to operate synchronously
We had an old PR here that was trying to do this, but got closed & not merged.   At any rate, the discussion on this PR is directly relevant to this ticket: https://github.com/neo4j-contrib/neo4j-streams/pull/161

currently, streams.publish takes 3 arguments:  topic, payload, and options.

At present, "options" basically gets disregarded and not used, but we need an option such as `{ synchronous: true }` which causes CALL streams.publish to BLOCK until the message has been delivered to kafka.   This is useful in several scenarios.  One is "chaining cypher statements" https://github.com/neo4j-contrib/neo4j-streams/issues/156 and the other is nested transactions, i.e. in a java plugin a user does this:

```
result = tx.run(query, statementParams);
``` 

inside of an update transaction, and needs to know whether to mark the outer TX as success or failure.

vga91:
Fixes neo4j-contrib#349
Fixes neo4j-contrib#158

  - Added `streams.publish.sync` procedure
  - Added `sendEventsSync` in `StreamsEventRouter.kt`
  - Added `throw RuntimeException("Payload may not be null")` in `streams.publish` and `streams.publish.sync` procedures

### DIFF
--- a/doc/asciidoc/procedures/index.adoc
+++ b/doc/asciidoc/procedures/index.adoc
@@ -91,6 +91,27 @@ Uses:
 
 This procedure return a `RecordMetadata` value like this `{"timestamp": 1, "offset": 2, "partition", 3, "keySize", 4, "valueSize", 5}`
 
+[cols="2*",options="header"]
+|===
+|Variable Name
+|Description
+
+|`timestamp`
+|The timestamp of the record in the topic/partition.
+
+|`offset`
+|The offset of the record in the topic/partition.
+
+|`partition`
+|The partition the record was sent to
+
+|`keySize`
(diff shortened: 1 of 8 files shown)


moxious (approved):
I reviewed the companion 4.0 PR here: https://github.com/neo4j-contrib/neo4j-streams/pull/379#pullrequestreview-550191167

Same comments apply; on this review I'm mostly looking for differences between 4.0 & 3.5 and I didn't see anything concerning.

PR #379, opened 2020-11-30
fixes #349: Provide option for streams.publish to operate synchronously (version 4.0)

### LINKED ISSUE #158: streams.produce doesn't publish in case of null argument

### LINKED ISSUE #349: Provide option for streams.publish to operate synchronously

vga91:
Fixes #349
Fixes #158

### DIFF
--- a/doc/asciidoc/procedures/index.adoc
+++ b/doc/asciidoc/procedures/index.adoc
@@ -117,6 +117,38 @@ You can send any kind of data in the payload, nodes, relationships, paths, lists
 
 In case of nodes or relationships if the topic is defined in the patterns provided by the configuration their properties will be filtered in according with the configuration.
 
+
+=== streams.publish.sync
+
+Similar to `streams.publish` procedure, but in a synchronous way.
+
+Uses:
+
+`CALL streams.publish.sync('my-topic', 'my-payload', {<config>}) YIELD value RETURN value`
+
+This procedure return a `RecordMetadata` value like this `{"timestamp": 1, "offset": 2, "partition", 3, "keySize", 4, "valueSize", 5}`
+
+[cols="2*",options="header"]
+|===
+|Variable Name
+|Description
+
+|`timestamp`
+|The timestamp of the record in the topic/partition.
+
+|`offset`
+|The offset of the record in the topic/partition.
+
+|`partition`
+|The partition the record was sent to
+
+|`keySize`
+|The size of the serialized, uncompressed key in bytes
+
+|`valueSize`
+|The size of the serialized, uncompressed value in bytes
+|===
+
 === streams.consume
 
 This procedure allows to consume messages from a given topic.
--- a/producer/src/main/kotlin/streams/kafka/KafkaEventRouter.kt
+++ b/producer/src/main/kotlin/streams/kafka/KafkaEventRouter.kt
@@ -2,6 +2,7 @@ package streams.kafka
 
 import org.apache.kafka.clients.producer.KafkaProducer
 import org.apache.kafka.clients.producer.ProducerRecord
+import org.apache.kafka.clients.producer.RecordMetadata
 import org.apache.kafka.common.KafkaException
 import org.apache.kafka.common.errors.AuthorizationException
 import org.apache.kafka.common.errors.OutOfOrderSequenceException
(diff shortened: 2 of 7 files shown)


doc/asciidoc/procedures/index.adoc
```diff
@@ -117,6 +117,38 @@ You can send any kind of data in the payload, nodes, relationships, paths, lists
 
 In case of nodes or relationships if the topic is defined in the patterns provided by the configuration their properties will be filtered in according with the configuration.
 
+
+=== streams.publish.sync
```
moxious, line 121: Good & specific docs

producer/src/main/kotlin/streams/kafka/KafkaEventRouter.kt
```diff
@@ -57,34 +59,40 @@ class KafkaEventRouter: StreamsEventRouter {
...
+                    log.debug("Successfully sent record in partition ${meta?.partition()} offset ${meta?.offset()} data ${meta?.topic()} key size ${meta?.serializedKeySize()}")
+                }
+                if (error != null) {
+                    if (log.isDebugEnabled) {
+                        log.debug("Error while sending record to ${producerRecord.topic()}, because of the following exception:", error)
+                    }
+                    // TODO add logging system here
                 }
             }
+            null
```
moxious, line 84: This is a minor point, that does not require change, but sometimes javadocs or the kotlin equivalent is nice in this condition.  The function looks like it returns Map<String,Any> which you would think is the result of publishing.  But it doesn't really -- the return type of the function is only relevant if sync: true.  The caller must know when to "take the return value seriously", which in our case the caller does, so it's ok

producer/src/main/kotlin/streams/kafka/KafkaEventRouter.kt
```diff
@@ -96,6 +104,17 @@ class KafkaEventRouter: StreamsEventRouter {
         send(producerRecord)
     }
 
+    override fun sendEventsSync(topic: String, transactionEvents: List<out StreamsEvent>): List<Map<String, Any>> {
+        producer.beginTransaction()
+
+        val results = transactionEvents.mapNotNull {
+            sendEvent(ThreadLocalRandom.current().nextInt(kafkaConfig.numPartitions), topic, it, true)
+        }
+        producer.commitTransaction()
```
moxious, line 113: +1 good for transaction handling

producer/src/main/kotlin/streams/procedures/StreamsProcedures.kt
```diff
@@ -22,31 +25,34 @@ class StreamsProcedures {
 
     @JvmField @Context var log: Log? = null
 
+    @Procedure(mode = Mode.READ, name = "streams.publish.sync")
+    @Description("streams.publish.sync(topic, payload, config) - Allows custom synchronous streaming from Neo4j to the configured stream environment")
+    fun sync(@Name("topic") topic: String?, @Name("payload") payload: Any?,
+             @Name(value = "config", defaultValue = "{}") config: Map<String, Any>?): Stream<StreamPublishResult> {
+        checkEnabled()
+        if (isTopicNullOrEmpty(topic)) {
+            return Stream.empty()
```
moxious, line 34: I'm ok with this for now but I can imagine users will ask for this to throw an error instead.

Imagine this use case:

```
MATCH (k:KafkaConfig { id: 1 })
MATCH (p:Person)
WITH k.topic as topic, collect(p.name) as names
CALL streams.publish.sync(topic, names, {})
```

Now imagine that k.topic = ''.    This will *look like it did something* but will actually do nothing at all; in this case the user needs to inspect the output to see if it's non-empty to know if they succeeded.   But it isn't clear which error semantics are best here, so we continue as it is for now.

moxious (approved):

PR #394, opened 2020-12-15
fixes #358: Dynamic Configuration

### LINKED ISSUE #358: Dynamic Configuration
This is an epic, or sort of meta-ticket, to sweep a lot of related tickets together.

The core underlying issue is that neo4j-streams is configured via neo4j.conf, which is only re-read when the database restarts. As a result of using the Neo4j config system, to make most changes to streams you need to restart Neo4j.  This is unacceptable in many scenarios since it might require cycling a cluster, for example.

This ticket is for brainstorming & discussion on best paths forward, keeping in mind several facets:

* Need to support 3.5 & 4.0 going forward
* The general need for dynamic plugin configuration

conker84:
Fixes #358

**[READ THIS DOCUMENTATION ABOUT THIS PR](https://docs.google.com/document/d/1ZbXOA89HBO2fATHJGc-A6gD4x7bkSfo_Ywet2X6GTYs/edit?usp=sharing)**

This PR adds new features in order to support configuration changes at runtime without the need of restarting the database each time, but simply by changing the new `streams.conf` or by provided procedures. 

  - Added new procedures `streams.configuration.set` and `streams.configuration.get`

Todo:
- [x] Fix broken tests
- [x] Add documentation

### DIFF
--- a/common/src/main/kotlin/streams/configuration/StreamsConfig.kt
+++ b/common/src/main/kotlin/streams/configuration/StreamsConfig.kt
@@ -1,0 +1,78 @@
+package streams.configuration
+
+import org.neo4j.kernel.internal.GraphDatabaseAPI
+import org.neo4j.logging.Log
+import org.neo4j.logging.internal.LogService
+import org.neo4j.plugin.configuration.ConfigurationLifecycle
+import org.neo4j.plugin.configuration.ConfigurationLifecycleUtils
+import org.neo4j.plugin.configuration.EventType
+import org.neo4j.plugin.configuration.listners.ConfigurationLifecycleListener
+import streams.utils.StreamsUtils
+import java.io.File
+import java.util.concurrent.ConcurrentHashMap
+
+class StreamsConfig(triggerPeriod: Int = DEFAULT_TRIGGER_PERIOD, log: Log) {
+
+    companion object {
+        @JvmStatic private val cache = ConcurrentHashMap<String, StreamsConfig>()
+
+        private const val DEFAULT_TRIGGER_PERIOD: Int = 10000
+
+        private const val SUN_JAVA_COMMAND = "sun.java.command"
+        private const val CONF_DIR_ARG = "config-dir="
+        private const val DEFAULT_PATH = "."
+
+        fun getNeo4jConfFolder(): String { // sun.java.command=com.neo4j.server.enterprise.CommercialEntryPoint --home-dir=/home/myid/neo4j-enterprise-4.0.0-alpha09mr02 --config-dir=/home/myid/neo4j-enterprise-4.0.0-alpha09mr02/conf
+            val command = System.getProperty(SUN_JAVA_COMMAND, "")
+            return command.split("--")
+                    .map(String::trim)
+                    .filter { it.startsWith(CONF_DIR_ARG) }
+                    .map { it.substring(CONF_DIR_ARG.length) }
+                    .firstOrNull() ?: DEFAULT_PATH
+        }
+
+        fun getInstance(db: GraphDatabaseAPI): StreamsConfig = cache.computeIfAbsent(StreamsUtils.getName(db)) {
+            StreamsConfig(log = db.dependencyResolver
+                    .resolveDependency(LogService::class.java)
+                    .getUserLog(StreamsConfig::class.java))
+        }
+
+        fun removeInstance(db: GraphDatabaseAPI): StreamsConfig? = cache.remove(StreamsUtils.getName(db))
+    }
+
+    private val configLifecycle: ConfigurationLifecycle
+
+    init {
+        val neo4jConfFolder = System.getenv().getOrDefault("NEO4J_CONF", getNeo4jConfFolder())
+        configLifecycle = ConfigurationLifecycle(triggerPeriod,
+                "$neo4jConfFolder${File.separator}streams.conf",
+                true, log, true, "streams.", "kafka.")
+    }
+
+    fun setProperty(key: String, value: Any, save: Boolean = true) {
+        configLifecycle.setProperty(key, value, save)
+    }
+
+    fun setProperties(map: Map<String, Any>, save: Boolean = true) {
+        configLifecycle.setProperties(map, save)
+    }
+
+    fun reload() {
+        configLifecycle.reload()
+    }
+
+    fun start() {
+        configLifecycle.start()
+    }
+
+    fun getConfiguration(): MutableMap<String, Any> = ConfigurationLifecycleUtils.toMap(configLifecycle.configuration)
+
+    fun addConfigurationLifecycleListener(evt: EventType,
+                                          listener: ConfigurationLifecycleListener) {
+        configLifecycle.addConfigurationLifecycleListener(evt, listener)
+    }
+
+    fun stop(shutdown: Boolean = false) {
+        configLifecycle.stop(shutdown)
+    }
+}
--- a/common/src/main/kotlin/streams/configuration/StreamsConfig.kt
+++ b/common/src/main/kotlin/streams/configuration/StreamsConfig.kt
@@ -57,6 +57,14 @@ class StreamsConfig(triggerPeriod: Int = DEFAULT_TRIGGER_PERIOD, log: Log) {
         configLifecycle.setProperties(map, save)
     }
 
+    fun removeProperty(key: String, save: Boolean = true) {
+        configLifecycle.removeProperty(key, save)
+    }
+
+    fun removeProperties(keys: Collection<String>, save: Boolean = true) {
+        configLifecycle.removeProperties(keys, save)
+    }
+
     fun reload() {
         configLifecycle.reload()
     }
--- a/common/src/main/kotlin/streams/configuration/StreamsConfigProcedures.kt
+++ b/common/src/main/kotlin/streams/configuration/StreamsConfigProcedures.kt
@@ -7,6 +7,13 @@ import org.neo4j.procedure.*
 import streams.events.KeyValueResult
 import java.util.stream.Stream
 
+data class StreamsConfigProceduresConfiguration(val save: Boolean) {
+    constructor(map: Map<String, Any>?): this(map.orEmpty()
+        .getOrDefault("save", "true")
+        .toString()
+        .toBoolean())
+}
+
 class StreamsConfigProcedures {
 
     @JvmField @Context
@@ -27,11 +34,24 @@ class StreamsConfigProcedures {
         }
         val map = properties.mapValues { it.value.toString() }
         val instance = StreamsConfig.getInstance(db as GraphDatabaseAPI)
-        val save = config.orEmpty()
-                .getOrDefault("save", "true")
-                .toString()
-                .toBoolean()
-        instance.setProperties(map, save)
+        val cfg = StreamsConfigProceduresConfiguration(config)
+        instance.setProperties(map, cfg.save)
+        return get()
+    }
+
+    @Admin
+    @Procedure
+    @Description("""
+        streams.configuration.remove(<properties_list>, <config_map>) YIELD name, value
+    """)
+    fun remove(@Name(value = "keys") properties: List<String>,
+            @Name(value = "config", defaultValue = "{}") config: Map<String, Any>?): Stream<KeyValueResult> {
+        if (properties.isNullOrEmpty()) {
+            throw RuntimeException("Property must be not empty")
+        }
+        val instance = StreamsConfig.getInstance(db as GraphDatabaseAPI)
+        val cfg = StreamsConfigProceduresConfiguration(config)
+        instance.removeProperties(properties, cfg.save)
         return get()
     }
 
--- a/common/src/main/kotlin/streams/configuration/StreamsConfigProcedures.kt
+++ b/common/src/main/kotlin/streams/configuration/StreamsConfigProcedures.kt
@@ -1,0 +1,48 @@
+package streams.configuration
+
+import org.neo4j.graphdb.GraphDatabaseService
+import org.neo4j.kernel.internal.GraphDatabaseAPI
+import org.neo4j.logging.Log
+import org.neo4j.procedure.*
+import streams.events.KeyValueResult
+import java.util.stream.Stream
+
+class StreamsConfigProcedures {
+
+    @JvmField @Context
+    var log: Log? = null
+
+    @JvmField @Context
+    var db: GraphDatabaseService? = null
+
+    @Admin
+    @Procedure
+    @Description("""
+        streams.configuration.set(<properties_map>, <config_map>) YIELD name, value
+    """)
+    fun set(@Name(value = "properties") properties: Map<String, Any>?,
+                @Name(value = "config", defaultValue = "{}") config: Map<String, Any>?): Stream<KeyValueResult> {
+        if (properties.isNullOrEmpty()) {
+            throw RuntimeException("Property must be not empty")
+        }
+        val map = properties.mapValues { it.value.toString() }
+        val instance = StreamsConfig.getInstance(db as GraphDatabaseAPI)
+        val save = config.orEmpty()
+                .getOrDefault("save", "true")
+                .toString()
+                .toBoolean()
+        instance.setProperties(map, save)
+        return get()
+    }
+
+    @Admin
+    @Procedure
+    @Description("""
+        streams.configuration.get() YIELD name, value
+    """)
+    fun get(): Stream<KeyValueResult> = StreamsConfig.getInstance(db as GraphDatabaseAPI)
+            .getConfiguration()
+            .entries
+            .map { KeyValueResult(it.key, it.value) }
+            .stream()
+}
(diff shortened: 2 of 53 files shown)


conker84:
[Uploading neo4j-streams-3.5.13-beta.jar.zip…]()
Added the `jar` in order to allow the to easily test the PR

common/src/test/kotlin/streams/configuration/StreamsConfigProceduresIT.kt
```diff
@@ -0,0 +1,41 @@
...
+    @After
+    fun tearDown() {
+        db.shutdown()
+    }
+
+    @Test
+    fun `should set properties`() {
+        db.dependencyResolver.resolveDependency(Procedures::class.java)
+                .registerProcedure(StreamsConfigProcedures::class.java, true)
+        val props = mapOf("streams.procedures.enabled" to "true")
```
moxious, line 33: is it possible to use the dynamic config procedure to set procedures.enabled=false, and thereby break yourself and be unable to change it because the procedures are no longer available?   This would be bad.

common/src/main/kotlin/streams/configuration/StreamsConfig.kt
```diff
@@ -0,0 +1,77 @@
...
+
+        fun removeInstance(db: GraphDatabaseAPI): StreamsConfig? = cache.remove(StreamsUtils.getName(db))
+    }
+
+    private val configLifecycle: ConfigurationLifecycle
+
+    init {
+        val neo4jConfFolder = System.getenv().getOrDefault("NEO4J_CONF", getNeo4jConfFolder())
+        configLifecycle = ConfigurationLifecycle(triggerPeriod,
+                "$neo4jConfFolder/streams.conf",
```
moxious, line 47: file path concat - use File separator for cross-platform

common/src/main/kotlin/streams/configuration/StreamsConfigProcedures.kt
```diff
@@ -0,0 +1,50 @@
...
+    """)
+    fun set(@Name(value = "properties") properties: Map<String, Any>?,
+                @Name(value = "config", defaultValue = "{}") config: Map<String, Any>?): Stream<KeyValueResult> {
+        if (properties.isNullOrEmpty()) {
+            throw RuntimeException("Property must be not empty")
+        }
+        val map = properties.mapValues { it.value.toString() }
+        val instance = StreamsConfig.getInstance(db as GraphDatabaseAPI)
+        val save = config.orEmpty()
+                .getOrDefault("save", "true")
```
moxious, line 34: This implies a unit test we should have; if I set "save": "false", I'd like to see that the config file doesn't change.   This would be useful for testing iterative changes without modifying the config maybe?

consumer/src/main/kotlin/streams/StreamsSinkConfigurationListner.kt
```diff
@@ -0,0 +1,146 @@
...
+    private val streamsTopicService = StreamsTopicService()
+
+    private fun shutdown() {
+        val isShuttingDown = eventSink != null
+        if (isShuttingDown) {
+            log.info("[Sink] Shutting down the Streams Sink Module")
+        }
+        eventSink?.stop()
+        eventSink = null
+        StreamsSinkProcedures.unregisterStreamsEventSink(db)
```
moxious, line 36: I understand this only shuts down the SINK.    What about the SOURCE?

What if I configure those dynamically?

Additionally -- if we do make the source dynamic in the config as well (which we should) then we need to consider this scenario:

1. User is producing -> Kafka
2. User changes some sink parameter (fine)
3. During the restart of the plugin, a TX comes through.
4. Did we miss producing the TX because we were stopped while it happened?  Or do we block the TX until we're ready to produce it out?

doc/asciidoc/overview/configuration.adoc
```diff
@@ -0,0 +1,200 @@
+[#neo4j_configuration_system]
+== Configuration system overview
+
+[abstract]
+--
+This documentation is valid for Neo4j Streams from versions `3.5.13`, prior versions
+works statically with the properties provided inside the `neo4j.conf` file
+--
+
+Since the version `3.5.13` we introduced a new configuration systems based on
```
moxious, line 10: Make a mini header "Location of Configuration Information", and say it both ways.

For versions prior to blah blah, use neo4j.conf.  For versions as of an after blah blah, use streams.conf.  

Make it super simple.

doc/asciidoc/overview/configuration.adoc
```diff
@@ -0,0 +1,200 @@
...
+a new `streams.conf` file inside `$NEO4J_HOME/conf` and put in there all the
+required configuration.
+
+==== Note about usage with Docker
+
+The official Neo4j Docker image uses a particular naming convention for environment
+variables in order to transform them into properties inside the `neo4j.conf` file.
+In order to be compliant with that behaviour you can still use them without changing anything
+in your configuration, under-the-hood the new configuration system
+will save them inside the `streams.conf` file instead.
```
moxious, line 25: As of version blah blah

doc/asciidoc/overview/configuration.adoc
```diff
@@ -0,0 +1,200 @@
...
+the reload of the plugins so you must use this feature very carefully.*
+
+=== File based changes
+
+Every change inside the `streams.conf` is gathered and reloads the Streams `Sink`
+and `Source` with the new configuration.
+
+=== Procedure based changes
+
+With the new configuration system you'll find two new procedures:
```
moxious, line 44: don't say new configuration system.  Say "Versions after blah blah"

doc/asciidoc/overview/configuration.adoc
```diff
@@ -0,0 +1,200 @@
...
+[cols="2*",options="header"]
+|===
+|Variable Name
+|Description
+
+|`plugin_config_map`
+|This map represents the set of configurations applied to the `Sink` and the `Source`
+
+|`procedure_config`
+|The configuration value
```
moxious, line 127: Is there anything I can say in the second map other than save: true/false?   If this ties to a map in Apache Commons or something like that, then link it.

doc/asciidoc/overview/configuration.adoc
```diff
@@ -0,0 +1,200 @@
...
+
+|"streams.sink.errors.log.enable"
+|"true"
+
+|"kafka.zookeeper.connect"
+|"zookeeper:2181"
+
+|===
+
+==== `streams.configuration.set`
```
moxious, line 111: You need a big fat NOTE/WARNING here about the side effects of doing this.  Users will assume that they're writing stuff to a file, OK fine.

But describe what the lifecycle consequences are here.  For example, "All existing sinks will be stopped & restarted, even if you only change one parameter" (for example, if that is true).

What happens to sources?

How long do we expect the consumption downtime to be?   

Can this fail?  If the restart post config change fails, how would we see / know?

common/src/test/kotlin/streams/configuration/StreamsConfigProceduresIT.kt
conker84: no this cannot happen because `streams.procedures.enabled` does not affect `streams.configuration.get/set`

common/src/main/kotlin/streams/configuration/StreamsConfig.kt
conker84: done thank you!

common/src/main/kotlin/streams/configuration/StreamsConfigProcedures.kt
conker84: I added one in the plugin lifecycle manager. Wdyt?

consumer/src/main/kotlin/streams/StreamsSinkConfigurationListner.kt
conker84: It shouts down both, please look at this class:
https://github.com/neo4j-contrib/neo4j-streams/pull/394/files#diff-4b0c66acb7d47e7d66f7c655ffffc1672fbc070f9058880bf0c7bfdc5a8715d5R14

Yes at this very moment we restart both but we can add a mechanism in order to check differences in order to check if source/sink properties are changed and then restart the module only if there is a property change for that module.

conker84: When the source is restarted we unplug the transaction handler so the database keeps working but we lost what it's doing during the restart

doc/asciidoc/overview/configuration.adoc
conker84: done

doc/asciidoc/overview/configuration.adoc
conker84: done

doc/asciidoc/overview/configuration.adoc
conker84: done

doc/asciidoc/overview/configuration.adoc
conker84: At this moment we use only that propertyy

common/src/main/kotlin/streams/configuration/StreamsConfigProcedures.kt
```diff
@@ -0,0 +1,50 @@
...
+                .toString()
+                .toBoolean()
+        instance.setProperties(map, save)
+        return get()
+    }
+
+    @Procedure(mode = Mode.READ)
```
moxious, line 41: Security -- need admin only annotation, or otherwise, we need to add docs indicating how people can restrict the usage of this procedure.  The config may include secret tokens & keys to auth to kafka that regular users should not be able to obtain.

common/src/main/kotlin/streams/configuration/StreamsConfigProcedures.kt
```diff
@@ -0,0 +1,50 @@
...
+
+class StreamsConfigProcedures {
+
+    @JvmField @Context
+    var log: Log? = null
+
+    @JvmField @Context
+    var db: GraphDatabaseService? = null
+
+    @Procedure
```
moxious, line 22: Security -- need admin only annotation, or otherwise, we need to add docs indicating how people can restrict the usage of this procedure.  The config may include secret tokens & keys to auth to kafka that regular users should not be able to obtain.

doc/asciidoc/overview/configuration.adoc
conker84: Done.

I added a mechanism that checks if source/sink properties are changed and restart the modules by that check:

_**N.b.** The Source/Sink module will be restarted only if there are changes in the
configuration related to itself; this means that if you have both active and
you change properties related to the Sink, only it will be restarted._

common/src/main/kotlin/streams/configuration/StreamsConfigProcedures.kt
conker84: Done

common/src/main/kotlin/streams/configuration/StreamsConfigProcedures.kt
conker84: Done

conker84:
@moxious thank you so much for the review! I updated the code-base please look at my comments.

conker84:
[neo4j-streams-3.5.13-beta1.jar.zip](https://github.com/neo4j-contrib/neo4j-streams/files/5809857/neo4j-streams-3.5.13-beta1.jar.zip)
This is the new jar with the last changes

txwikinger (changes requested):
When a sink is created that has a syntax error in the CYPHER statement, or an existing sink is overridden in such a way, even after changing it back, the sink does not work.

There are no errors in the logs. It is really difficult to figure out why nothing works anymore. The changes are correctly put into the configuration file and correctly appear when queried via streams.configuration.get()

Also, neither key or value for setting a configuration are in any way validated. I am not sure if this is very easy to do, however, at minimum, there should be some kind of errors in the logs at the beginning (need to see what is needed for Aura) that indicate when either the key or value create an error conditions. In addition, a key cannot be deleted, hence a wrongly formed key requires the restart of the database after manually editing the configuration file (this invalidates the notion of dynamic configuration)

conker84:
> When a sink is created that has a syntax error in the CYPHER statement, or an existing sink is overridden in such a way, even after changing it back, the sink does not work.

I can add a Cypher validation but when I put the valid cypher statement back my configuration started to consume again the messages, can you detail the steps that you did?
I tried in this way:
* Started with a valid cypher
* Changed to an invalid query -> consumed some messages -> and I got this error into the log:
```bash
ERROR Error processing 1 messages
core1        | ErrorData(originalTopic=mytopic, timestamp=1611660781335, partition=0, offset=0, exception=org.neo4j.graphdb.QueryExecutionException: Invalid input 'f': expected whitespace, comment, '(', '.', node labels, '[', "=~", IN, STARTS, ENDS, CONTAINS, IS, '^', '*', '/', '%', '+', '-', '=', '~', "<>", "!=", '<', '>', "<=", ">=", AND, XOR, OR, ',' or '}' (line 1, column 84 (offset: 83))
core1        | "UNWIND {events} AS event CREATE (p:Person{name: event.name, surname: event.surname from: 'JSON'})"
core1        |                                                                                     ^, key=null, value={"name":"Andrea","surname":"Santurbano"}, executingClass=class streams.kafka.KafkaAutoCommitEventConsumer) Error processing 1 messages
core1        | ErrorData(originalTopic=mytopic, timestamp=1611660781335, partition=0, offset=0, exception=org.neo4j.graphdb.QueryExecutionException: Invalid input 'f': expected whitespace, comment, '(', '.', node labels, '[', "=~", IN, STARTS, ENDS, CONTAINS, IS, '^', '*', '/', '%', '+', '-', '=', '~', "<>", "!=", '<', '>', "<=", ">=", AND, XOR, OR, ',' or '}' (line 1, column 84 (offset: 83))
core1        | "UNWIND {events} AS event CREATE (p:Person{name: event.name, surname: event.surname from: 'JSON'})"
core1        |                                                                                     ^, key=null, value={"name":"Andrea","surname":"Santurbano"}, executingClass=class streams.kafka.KafkaAutoCommitEventConsumer)
core1        | streams.service.errors.ProcessingError: Error processing 1 messages
core1        | ErrorData(originalTopic=mytopic, timestamp=1611660781335, partition=0, offset=0, exception=org.neo4j.graphdb.QueryExecutionException: Invalid input 'f': expected whitespace, comment, '(', '.', node labels, '[', "=~", IN, STARTS, ENDS, CONTAINS, IS, '^', '*', '/', '%', '+', '-', '=', '~', "<>", "!=", '<', '>', "<=", ">=", AND, XOR, OR, ',' or '}' (line 1, column 84 (offset: 83))
core1        | "UNWIND {events} AS event CREATE (p:Person{name: event.name, surname: event.surname from: 'JSON'})"
core1        |                                                                                     ^, key=null, value={"name":"Andrea","surname":"Santurbano"}, executingClass=class streams.kafka.KafkaAutoCommitEventConsumer)
core1        |  at streams.service.errors.KafkaErrorService.report(KafkaErrorService.kt:37)
core1        |  at streams.kafka.KafkaAutoCommitEventConsumer.executeAction(KafkaAutoCommitEventConsumer.kt:88)
core1        |  at streams.kafka.KafkaAutoCommitEventConsumer.readSimple(KafkaAutoCommitEventConsumer.kt:80)
core1        |  at streams.kafka.KafkaAutoCommitEventConsumer.read(KafkaAutoCommitEventConsumer.kt:109)
core1        |  at streams.kafka.KafkaEventSink$createJob$1.invokeSuspend(KafkaEventSink.kt:125)
core1        |  at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
core1        |  at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:56)
core1        |  at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:571)
core1        |  at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:738)
core1        |  at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:678)
core1        |  at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:665)
```
* Restored the old valid Cypher query the messages started to be consumed again

> There are no errors in the logs. It is really difficult to figure out why nothing works anymore. The changes are correctly put into the configuration file and correctly appear when queried via streams.configuration.get()
> 
> Also, neither key or value for setting a configuration are in any way validated. I am not sure if this is very easy to do, however, at minimum, there should be some kind of errors in the logs at the beginning (need to see what is needed for Aura) that indicate when either the key or value create an error conditions. In addition, a key cannot be deleted, hence a wrongly formed key requires the restart of the database after manually editing the configuration file (this invalidates the notion of dynamic configuration)

We have a minimum set of validation on keys can detail with a simple use-case?

Thank you so much for testing this @txwikinger !!!

txwikinger (approved):

PR #405, opened 2021-01-18
Fixes #272: Allow to publish messages with key in Kafka topic

### LINKED ISSUE #272: Allow to publish messages with key in Kafka topic
## Allow to publish messages with key in Kafka topic.
Using streams.publish procedure add feature which will allow to publish message with key in Kafka topic. By adding this new feature we will be able to leverage all the advantages of having key in topic partitioning, order, log compaction, we can use the topic in joining multiple streams etc.

## Considered alternatives
As a work around we can write our own Kafka producer which will publish the messages based on the result of the cypher query. 

By adding this new feature we will be able to leverage all the advantages of having key in topic such as partitioning, order, log compaction, we can use the topic in joining multiple streams etc..

vga91:
Fixes #272

### DIFF
--- a/doc/docs/modules/ROOT/pages/procedures.adoc
+++ b/doc/docs/modules/ROOT/pages/procedures.adoc
@@ -113,6 +113,23 @@ Input Parameters:
 
 |===
 
+Configuration parameters:
+[cols="3*",options="header"]
+|===
+|Name
+|Type
+|Description
+
+|`key`
+|Object
+|The key of message that you want to stream
+
+|`partition`
+|Int
+|The partition of message that you want to stream
+
+|===
+
 You can send any kind of data in the payload, nodes, relationships, paths, lists, maps, scalar values and nested versions thereof.
 
 In case of nodes or relationships if the topic is defined in the patterns provided by the configuration their properties will be filtered in according with the configuration.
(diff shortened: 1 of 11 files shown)


doc/docs/modules/ROOT/pages/procedures.adoc
```diff
@@ -113,6 +113,23 @@ Input Parameters:
...
+Configuration parameters:
+[cols="3*",options="header"]
+|===
+|Name
+|Type
+|Description
+
+|`key`
+|Object
+|The key of message that you want to stream
```
moxious, line 125: actual behavior is that if the key specified doesn't exist, you get a random UUID and no warning, so you should note this right here.   Is this the correct behavior?  Note that if someone has a tiny typo in what they provide to the config, it might cause problems.  Imagine you have records with `customerId` and you say that `customerid` is the key.   Should this fail, or produce a random UUID which could cause problems downstream?

moxious (commented):
code is generally ok -- just one question about behavior

doc/docs/modules/ROOT/pages/procedures.adoc
vga91: I updated the documentation with the random UUID note. 
Moreover, I have modified "The key of message" with the most appropriate "The **value key** of message".
Your observation makes me think about a potential bug with  null key, so I've updated the code for this use case. 
Now if I have a query like `"CREATE (c:Customer {id :1}) WITH c CALL streams.publish('customers', c, {key: c.foo})"` the message is also sent with null key, and you use the partitioning system defined in the broker.

moxious (approved):

conker84:
@vga91 thank you

PR #406, opened 2021-01-18
Fixes #305: CDC and Kafka Log Compaction

### LINKED ISSUE #305: CDC and Kafka Log Compaction

vga91:
Fixes #305


producer/src/main/kotlin/streams/Extensions.kt
```diff
@@ -41,4 +42,35 @@ fun ConstraintDefinition.isNodeConstraint(): Boolean {
...
+
+fun StreamsTransactionEvent.asSourceRecordValue(strategy: String): StreamsTransactionEvent? =
+        if(isStrategyCompact(strategy) && meta.operation == OperationType.deleted) null else this
+
+fun StreamsTransactionEvent.asSourceRecordKey(strategy: String): Any =
+        when {
+            strategy == TopicConfig.CLEANUP_POLICY_DELETE -> "${meta.txId + meta.txEventId}-${meta.txEventId}"
+            isStrategyCompact(strategy) && payload is NodePayload -> getKeyOfNode(payload as NodePayload, schema)
+            isStrategyCompact(strategy) && payload is RelationshipPayload -> getKeyOfRel(payload as RelationshipPayload)
+            else -> throw RuntimeException("Invalid kafka.streams.log.compaction.strategy value: $strategy")
```
moxious, line 55: This is a runtime error of a problem that should be detectable at configure time.

Can you please eliminate this thrown error message, which will spam every time something is sent over kafka (correct?)

Actual behavior should be that when you specify a bad config `kafka.streams.log.compaction.strategy=(╯°□°）╯︵ ┻━┻` that the plugin at start time should say "Invalid log compaction strategy setting, switching to default value XYZ" and then at runtime use the correct default.

doc/docs/modules/ROOT/pages/producer-configuration.adoc
```diff
@@ -45,6 +46,12 @@ You can tune the custom property `kafka.topic.discovery.polling.interval` in ord
 periodically check for new topics into the Kafka cluster so the plugin will be able
 to send events to the defined topics.
 
+
+With `kafka.streams.log.compaction.strategy=delete` will be generated a sequence of unique keys with Neo4j Streams Source.
+instead with `kafka.streams.log.compaction.strategy=compact` the keys will be adapted to enable
+https://kafka.apache.org/documentation.html#compaction[Log Compaction] on the Kafka side.
+Namely, the operations which will involve the same nodes or relationships, will have the same key.
```
moxious, line 53: State what the default is if you don't choose, which will be most users.    Line above suggests that it's delete?

Also, why do we call it delete?  This sounds like something's getting deleted....now when logs get compacted to 500mb (or whatever) that's probably true, things get deleted.  But maybe this should be called "unique" or something?  (Because the IDs end up unique and not compact-able) I'm not sure.  Which is the least confusing for users?

producer/src/main/kotlin/streams/kafka/KafkaConfiguration.kt
```diff
@@ -28,6 +29,7 @@ data class KafkaConfiguration(val zookeeperConnect: String = "localhost:2181",
                               val transactionalId: String = StringUtils.EMPTY,
                               val lingerMs: Int = 1,
                               val topicDiscoveryPollingInterval: Long = TimeUnit.MINUTES.toMillis(5),
+                              val streamsLogCompactionStrategy: String = TopicConfig.CLEANUP_POLICY_DELETE,
```
moxious, line 32: ah, there's the default.  It's fine to pick whichever default you wish (best to keep it the way it's been in the past for compatibility reasons).  See note earlier about naming / delete

producer/src/test/kotlin/streams/integrations/KafkaEventRouterLogCompactionTSE.kt
```diff
@@ -0,0 +1,627 @@
...
+        db.execute("CALL streams.publish('$topic', 'Compaction 2', {key: '$keyRecord'})")
+        db.execute("CALL streams.publish('$topic', 'Compaction 3', {key: 'Foo'})")
+        db.execute("CALL streams.publish('$topic', 'Compaction 4', {key: '$keyRecord'})")
+
+        // to activate the log compaction process we publish dummy messages
+        db.execute("UNWIND range(1,9999) as id CALL streams.publish('$topic', 'id', {key: 'id'}) RETURN null") {
+            assertEquals(9999, Iterators.count(it))
+        }
+        // wait for messages population in topic
+        Thread.sleep(15000)
```
moxious, line 72: your choice on how best to do this but my experience is whenever you have Thread.sleep in a test, the test ends up being flaky the one time kafka is too slow, or whatever.  Consider polling longer, until you get expected number of records.    This code pattern (Thread.sleep followed by poll) is repeated many times in these tests, you can refactor into a func to do it once in a more resilient way

producer/src/test/kotlin/streams/integrations/KafkaEventRouterLogCompactionTSE.kt
```diff
@@ -0,0 +1,627 @@
...
+        db.execute("""
+            |MATCH (pippo:Person {name:'Pippo'})
+            |MATCH (pluto:Person {name:'Pluto'})
+            |MERGE (pippo)-[:KNOWS]->(pluto)
+        """.trimMargin())
+        db.execute("MATCH (:Person {name:'Pippo'})-[rel:KNOWS]->(:Person {name:'Pluto'}) DELETE rel")
+
+        // to activate the log compaction process we create dummy messages
+        createManyPersons()
+        Thread.sleep(20000)
```
moxious, line 104: see poll note above

producer/src/test/kotlin/streams/integrations/KafkaEventRouterLogCompactionTSE.kt
```diff
@@ -0,0 +1,627 @@
...
+            |MERGE (pippo)-[:KNOWS]->(pluto)
+        """.trimMargin())
+
+        // we delete the relation, so will be created a tombstone record
+        db.execute("MATCH (:Person {name:'Pippo'})-[rel:KNOWS]->(:Person {name:'Pluto'}) DELETE rel")
+        db.execute("CREATE (:Person {name:'Paperino'})")
+
+        // to activate the log compaction process we create dummy messages
+        createManyPersons()
+        Thread.sleep(20000)
```
moxious, line 142: see poll note above

producer/src/test/kotlin/streams/integrations/KafkaEventRouterLogCompactionTSE.kt
```diff
@@ -0,0 +1,627 @@
...
+
+        db.execute("CREATE (:Person {name:'Watson'})")
+        db.execute("CREATE (:Person {name:'Sherlock'})")
+        db.execute("MATCH (p:Person {name:'Sherlock'}) SET p.address = '221B Baker Street'")
+        // we delete a node, so will be created a tombstone record
+        db.execute("MATCH (p:Person {name:'Sherlock'}) DETACH DELETE p")
+
+        // to activate the log compaction process we create dummy messages and we waiting for message population
+        createManyPersons()
+        Thread.sleep(20000)
```
moxious, line 171: see poll note above

producer/src/test/kotlin/streams/integrations/KafkaEventRouterLogCompactionTSE.kt
```diff
@@ -0,0 +1,627 @@
...
+        kafkaConsumer.subscribe(listOf(topic))
+
+        db.execute("CREATE (:Person {name:'Watson'})")
+        db.execute("CREATE (:Person {name:'Sherlock'})")
+        db.execute("MATCH (p:Person {name:'Sherlock'}) SET p.address = '221B Baker Street'")
+
+        // to activate the log compaction process we create dummy messages and we waiting for message population
+        db.execute("MATCH (p:Person {name:'Sherlock'}) DETACH DELETE p")
+        createManyPersons()
+        Thread.sleep(20000)
```
moxious, line 197: see poll note above

producer/src/test/kotlin/streams/integrations/KafkaEventRouterLogCompactionTSE.kt
```diff
@@ -0,0 +1,627 @@
...
+                mapOf("streams.source.topic.nodes.${topic}" to "Person{*}")
+        )
+        kafkaConsumer.subscribe(listOf(topic))
+
+        assertFailsWith(TransactionFailureException::class, "Invalid kafka.streams.log.compaction.strategy value: $invalid") {
+            db.execute("CREATE (:Person {name:'Pippo'})")
+        }
+    }
+
+}
```
moxious, line 627: really nice work with all of these tests 💯

moxious (changes requested):

producer/src/test/kotlin/streams/integrations/KafkaEventRouterLogCompactionTSE.kt
conker84: @vga91 please switch to `assertEventually`

producer/src/main/kotlin/streams/Extensions.kt
conker84: @vga91 please add a check into the `validate` method of `KafkaConfiguration`

doc/docs/modules/ROOT/pages/producer-configuration.adoc
vga91: I chose these two values to match the topic config `cleanup.policy` which has compact / delete values to enable log compaction or not. Of course, the term unique sounds good.
So, If you want, I can change `delete` to `unique`

moxious: I'm ok with the decision as it is, just make it clear in the docs.

vga91: Ok, i've added a note in the `.adoc`

doc/docs/modules/ROOT/pages/producer-configuration.adoc
```diff
@@ -45,6 +46,13 @@ You can tune the custom property `kafka.topic.discovery.polling.interval` in ord
 periodically check for new topics into the Kafka cluster so the plugin will be able
 to send events to the defined topics.
 
+
+With `kafka.streams.log.compaction.strategy=delete` will be generated a sequence of unique keys with Neo4j Streams Source.
+instead with `kafka.streams.log.compaction.strategy=compact` the keys will be adapted to enable
+https://kafka.apache.org/documentation.html#compaction[Log Compaction] on the Kafka side.
+Please note that delete strategy does not actually delete records, it has this name  to match the topic config `cleanup.policy=delete/compact`.
+Namely, the operations which will involve the same nodes or relationships, will have the same key.
+
```
conker84, line 55: Please add the docs regarding the key structure for nodes and relationships for both use-case with/without unique contraints

conker84 (changes requested):
Consider what we said in slack about improving the structure of the message keys.
Also add a test like `KafkaEventRouterEnterpriseTSE` that cover `NODE KEY` constraints

moxious:
@conker84 to review & later merge

conker84 (changes requested):

conker84 (approved):
LGTM thank you @vga91 !

PR #415, opened 2021-01-29
fixes #407: Move the internal procedures calls to direct API calls

### LINKED ISSUE #407: Move the internal procedures calls to direct API calls
i.e. `dbms.cluster.role()`

This will allow to reduce the amount of queries logged inside `query` log

N.b. we can do it only by reflection otherwise we have to include `enterprise` packages

conker84:
Fixes #407

Moved `dbms.cluster.role` invokation through API call via Reflection.

  - implemented the feature


moxious:
@conker84 to propose questions to clustering team, @moxious to follow up on those.

conker84:
@moxious In order to close this one I need to check the internals of `dbms.cluster.overview`

moxious:
@conker adding related issue: https://github.com/neo4j-contrib/neo4j-streams/issues/451

Note on that issue that polling & query log spam is expected (that's what this PR is addressing) but that ticket reports even more aggressive polling than should be the case on the 4.0 dynamic config branch.  Might be worth checking into.

Additionally, I asked @hugofirth to take a look at this PR and he'll be providing feedback when he can.

conker84:
I had a chat with @davidfauth and this fixed #451

common/src/main/kotlin/streams/utils/ProcedureUtils.kt
```diff
@@ -0,0 +1,71 @@
...
+
+    @JvmStatic
+    private val isLeaderMethodHandle = raftMachineClass?.let {
+        val lookup = MethodHandles.lookup()
+        lookup.findVirtual(it, "isLeader", MethodType.methodType(Boolean::class.java))
+            .asType(MethodType.methodType(Boolean::class.java, Any::class.java))
+    }
+
+    fun clusterMemberRole(db: GraphDatabaseAPI): String {
+        val fallback: (Exception?) -> String = { e: Exception? ->
```
hugofirth, line 30: I'm not certain why we use reflection to resolve the `RaftMachine` and then fall back on the `dbms.cluster.role` procedure. The latter is the supported mechanism and the implementation calls through to `RaftMachine#isLeader` same as this.

common/src/main/kotlin/streams/utils/ProcedureUtils.kt
```diff
@@ -0,0 +1,71 @@
...
+            e?.let { userLog.warn("Cannot call the APIs, trying with the cypher query", e) }
+                ?: userLog.warn("Cannot call the APIs, trying with the cypher query")
+            db.execute("CALL dbms.cluster.role(\$database)",
+                mapOf("database" to db.databaseName())
+            ) { it.columnAs<String>("role").next() }
+        }
+        val execute = {
+            raftMachineClass?.let {
+                try {
+                    val raftMachine: Any = db.dependencyResolver.resolveDependency(raftMachineClass)
```
hugofirth, line 43: If we must use the dependency resolver to fetch something internal I'd rather we use a less powerful interface, like `LeaderLocator`, which will give you the `RaftMachine` but without most of the API. If you want to use `LeaderLocator#getLeaderInfo` you'll also need to resolve the `ClusteringIdentityModule` in order to compare `LeaderLocator#memberId()` to `ClusteringIdentityModule#memberId( id for this database )`.

hugofirth (changes requested):
I have some concerns I'd like to talk through, wrt using reflection to resolve access to powerful internal classes.

I think the two best choices here are to stick with `dbms.cluster.role` or to work with `LeaderLocator#registerListener` and to create your own `LeaderListener` implementation.

common/src/main/kotlin/streams/utils/ProcedureUtils.kt
conker84: The fallback is for an _extrema ratio_ if something went wrong, in case of `UnsatisfiedDependencyException` so in case the class is not in there we manually manage it as normal case

common/src/main/kotlin/streams/utils/ProcedureUtils.kt
```diff
@@ -0,0 +1,71 @@
...
+                    if (isLeader) "LEADER" else "FOLLOWER"
+                } catch (e: UnsatisfiedDependencyException) {
+                    "LEADER"
+                }
+            } ?: "LEADER"
+        }
+        return executeOrFallback(execute, fallback)
+    }
+
+    fun isCluster(db: GraphDatabaseAPI): Boolean = try {
```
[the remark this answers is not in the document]
conker84: done

common/src/main/kotlin/streams/utils/ProcedureUtils.kt
conker84: Sorry Hugo but it don't have access to Neo4j codebase so it would be good to have more info via email if it's possibile

hugofirth: There is an interface which exists on all Neo4j versions from 3.5 on called `CoreMetaData`. 

The `RaftMachine` implements this interface. Its source is:

```
public interface CoreMetaData
{
    boolean isLeader();
}
```

If you use `DependencyResolver` to resolve `CoreMetaData` you will be getting the capability you need, without accessing `RaftMachine` itself 🙂

hugofirth (approved):

conker84:
Merging this as the Java CI is green

PR #431, opened 2021-02-24
Remove the temporary hack to build on fork

Mogztter:
Remove the temporary definition :sweat_smile:

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -4,8 +4,6 @@ on:
   push:
     branches:
       - '4.0'
-      # temporary: allow to see GH Actions result on fork
-      - '4-0-gh-actions'
   pull_request:
     branches:
       - '*'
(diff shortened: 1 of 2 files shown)


Mogztter:
Interesting, some links are broken.
I think they are all related to the "Edit this Page" link feature. I will take a closer look :eyes:

Mogztter:
So GitHub automatically sets `CI=true` and we are using this environment variable in the documentation UI to enable edit links: https://github.com/neo4j-documentation/docs-ui/blob/5a2b61262b143d887b0993d52959865b755838d8/src/partials/edit-this-page-link.hbs#L2

The problem is that edit links are using `HEAD` instead of the target branch `4.0`.
As a result, edit links cannot be resolved:

- :green_circle: OK: https://github.com/neo4j-contrib/neo4j-streams/edit/4.0/doc/docs/modules/ROOT/pages/procedures.adoc
- :red_circle: KO: https://github.com/neo4j-contrib/neo4j-streams/edit/HEAD/doc/docs/modules/ROOT/pages/procedures.adoc

Since edit links are automatically generated by Antora, we don't need to check them (at least not here).
I added two additional ignore rules to fix this issue.

PR #435, opened 2021-02-26
Fixes #304: : Random selection of partitions and order guarantees

### LINKED ISSUE #304: Random selection of partitions and order guarantees
Hi everyone !! This a request for improvements to asses some limitations in the partition selection strategy of the Kafka producer in case of multiple partitions per topic. Looking at the code, It seems that the partition is randomly selected from those available (**sendEvents** function in [KafkaEventRouter.kt](https://github.com/neo4j-contrib/neo4j-streams/blob/3.5/producer/src/main/kotlin/streams/kafka/KafkaEventRouter.kt)): 
```
producer.beginTransaction()
transactionEvents.forEach {
        val partition = ThreadLocalRandom.current().nextInt(kafkaConfig.numPartitions)
        if (it is StreamsTransactionEvent) {
            sendEvent(partition, topic, it)
      } else {
            sendEvent(partition, topic, it)
      }
}
producer.commitTransaction()
```
**Why is it a problem?** Partitions are the unit of parallelism and Kafka provide order guarantees for each of them: every partition is assigned to a single consumer per group and records within a partition are delivered in a FIFO fashion. Consider a scenario where I'm using Neo4j Streams Source to CDC a set of nodes. Using **n** partitions I can run up to a maximum of **n** consumer, load balancing the workload between them. **In this scenario, I may need order guarantees on events regarding a _single node_.** Since the Kafka producer randomly selects a partition, events about the same node are possibly stored in different partitions and processed by different consumers, without any order guarantees. From Kafka documentation:

> Kafka only provides a total order over records within a partition, not between different partitions in a topic. **Per-partition ordering combined with the ability to partition data by key is sufficient for most applications.** However, if you require a total order over records this can be achieved with a topic that has only one partition, though this will mean only one consumer process per consumer group. 

In my opinion, the user should be able to configure what order guarantees he wants: if total order is needed on the whole set of events, the selection of partitions does not matter, but the best thing you can do is have a topic with a single partition and only one consumer per group. **If order guarantees are required for events relating to the same node, the partition must be selected using the node id to allow the use of multiple consumers.** In the latter case, the above code could be modified as follows:
```
producer.beginTransaction()
transactionEvents.forEach {
        if (it is StreamsTransactionEvent) {
            val partition = it.payload.id.hashCode() % kafkaConfig.numPartitions
            sendEvent(partition, topic, it)
      } else {
            val partition = ThreadLocalRandom.current().nextInt(kafkaConfig.numPartitions)
            sendEvent(partition, topic, it)
      }
}
producer.commitTransaction()
```
Let me know what you think. Is there another way to get around this? Thanks !!

vga91:
Fixes #304


moxious:
rebase for docs then merge

conker84:
All green I'm merging this as we agreed

PR #440, opened 2021-03-04
Fixes #429: [CDC] Schema select randomly the uk in case of multiple unique constraints

### LINKED ISSUE #429: [CDC] Schema select randomly the uk in case of multiple unique constraints
We have the following structure in Neo4J:
(:Profile {uuid: "abc")-[:HAS_ACCESS_TO]->(:Vendor {uuid: "defg", id: 2})
Also, we have UNIQUE constraint for the Profile label "uuid" property. And we have two UNIQUE constraints for the Vendor label - one for "id" and another for "uuid" property.

This leads to an interesting situation. When we look at CDC events for HAS_ACCESS_TO relationship, we see different values for "event.payload.end.keys" (for Vendor label). Sometimes we see "id" there, and sometimes it's "uuid"

vga91:
Fixes #429

Sorting by `properties.size`, then by `label` name and finally by `properties` name alphabetically

### DIFF
--- a/common/src/main/kotlin/streams/utils/SchemaUtils.kt
+++ b/common/src/main/kotlin/streams/utils/SchemaUtils.kt
@@ -13,10 +13,9 @@ object SchemaUtils {
                             && propertyKeys.containsAll(constraint.properties)
                             && labels.contains(constraint.label)
                 }
-                .minBy { it.properties.size }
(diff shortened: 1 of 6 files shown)


conker84 (approved):
LGTM @vga91 thank you

PR #444, opened 2021-03-18
add `neo4j` prefix to `database` property name

alexwoolford:
The property should be `neo4j.database` and not just `database`.


conker84 (approved):
Thank you @alexwoolford

conker84:
The docs are broken because of #445, and will be fixed with #446. This is safe to merge

PR #447, opened 2021-03-19
Fixes #316: Required Kafka ACLs not documented

### LINKED ISSUE #316: Required Kafka ACLs not documented
On top of the expected READ/WRITE/CREATE on the configured topics, the sink requires access to list topics/configs on the Kafka cluster, for example.
Procedures don't have such access requirements. 

The documentation lacks information about what ACL access is required for the plugin to be used as sink/source/procedures, which makes using a secured broker a game of trial and error, and may result in suboptimal access control.

Documentation should list all Kafka ACL operations that need to be authorised for the client user, for sink/source/procedures.
Maybe logging could also been improved to indicate what operation was attempted when an authorization failure is reported.

Authorization is not documented.
Generic exception raised ("org.apache.kafka.common.errors.ClusterAuthorizationException: Cluster authorization failed.") when failing to authorise.

Create restrictive ACLs on Kafka that blocks DESCRIBE_CONFIGS for the user. That should cause calls to listTopics() to fail (among others).

### Steps (Mandatory)

  1. set up Kafka with authentication and authorization
  2. set up authentication parameters in plugin

### Versions

  - OS: *
  - Neo4j: 4.0.4
  - Neo4j-Streams: 4.0.1

mroiter-larus:
Fixes #316

### DIFF
--- a/doc/docs/modules/ROOT/pages/kafka-ssl.adoc
+++ b/doc/docs/modules/ROOT/pages/kafka-ssl.adoc
@@ -195,5 +195,83 @@ kafka.sasl.mechanism=PLAIN
 
 For more information, please consult the official Confluent documentation at the following links:
 
-* https://docs.confluent.io/2.0.0/kafka/sasl.html#authentication-using-sasl
-* https://docs.confluent.io/2.0.0/kafka/sasl.html#configuring-kafka-clients
+* https://docs.confluent.io/platform/current/kafka/authentication_sasl/index.html
+
+== Authorization with ACL's
+
+To configure use with ACLs, the following configuration properties are required:
+
+[source, properties]
+----
+kafka.authorizer.class.name=kafka.security.authorizer.AclAuthorizer
+kafka.zookeeper.set.acl=true
+----
+
+[NOTE]
+* `kafka.security.authorizer.AclAuthorizer` (the default Kafka authorizer implementation), was introduced in Apache Kafka 2.4/Confluent Platform 5.4.0. If you are running a previous version, then use SimpleAclAuthorizer (`kafka.security.auth.SimpleAclAuthorizer`). If you are using the Confluent platform, you can use also the LDAP authorizer (please refer to the official Confluent documentation for further details: https://docs.confluent.io/platform/current/security/ldap-authorizer/quickstart.html)
+* Please consider that `zookeeper.set.acl` is **false** by default
+
+From the official Kafka documentation you can find that if a resource has no associated ACLs, then no one is allowed to access that resource except super users.
+If this is the case in your Kafka cluster, then you have also to add the following:
+
+[source, properties]
+----
+kafka.allow.everyone.if.no.acl.found=true
+----
+
+[NOTE]
+Be very careful on using the above property because, as the property name implies, it will allow access to everyone if no acl were found
+
+If super users are specified, then include also:
+
+[source,properties]
+----
+kafka.super.users=...
+----
+
+Moreover, if you change the default user name (principal) mapping rule then you have to add also the following properties:
+
+* If you used SSL encryption, then:
+
++
+[source, properties]
+----
+kafka.ssl.principal.mapping.rules=...
+----
+
+* If you used SASL encryption (probably so, if you have Kerberos environment), then:
+
++
+[source, properties]
+----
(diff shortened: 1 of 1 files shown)


moxious:
Next step @moxious to ask FE to review to ensure this closes the issue

moxious:
@mroiter-larus feedback from issue submitter:

> So this describes how to configure the kafka client. My point was more about documenting which ACLs the plugin actually needs. Does it need to read/write/list topics? Does needs other ACLs to retrieve configuration data from the broker? And also a breakdown fore sink/source/procs, as each would have different needs, and can be enabled/disabled independently.

> So, it doesn’t really address the issue for me.

As far as I know we don't need to list topics.   We need read for sink, and write for source.  Do we ever need configuration data from the broker?  Not that I know of.

Can you revise to add some of this?   Also,  please rebase the docs fix, so we can get green PR on the next go-around.

mroiter-larus:
For sure we need a read ACL for the sink and a write ACL for the source. I think we don't need configuration data from the broker. The only configuration data we read, for example with the procedure `streams.configuration.get`, is coming from the `neo4j.conf`. I'll revise the PR

mroiter-larus:
@moxious PR updated. I hope this time i've covered the lacks in the documentation

doc/docs/modules/ROOT/pages/kafka-ssl.adoc
```diff
@@ -195,5 +195,79 @@ kafka.sasl.mechanism=PLAIN
 
 For more information, please consult the official Confluent documentation at the following links:
 
-* https://docs.confluent.io/2.0.0/kafka/sasl.html#authentication-using-sasl
-* https://docs.confluent.io/2.0.0/kafka/sasl.html#configuring-kafka-clients
+* https://docs.confluent.io/platform/current/kafka/authentication_sasl/index.html
+
+== Authorization with ACL's
+
+When Kafka is configured to use ACL's, you have to add also the following properties on the `neo4j.conf` file:
```
moxious, line 202: To configure use with ACLs, the following configuration properties are required.


(Note:  do not reference neo4j.conf because we're about to introduce dynamic config)

moxious: Also: does this section of documentation apply only to the plugin?  I think it does.  If so, just make a note.

doc/docs/modules/ROOT/pages/kafka-ssl.adoc
```diff
@@ -195,5 +195,79 @@ kafka.sasl.mechanism=PLAIN
...
+
+[source, properties]
+----
+kafka.security.inter.broker.protocol=SASL_SSL
+----
+
+[NOTE]
+The last property is `PLAIN` by default
+
+To make the plugin works properly, the operations that has to be authorized for the Topic and Cluster resource types are:
```
moxious, line 261: Revise:

"To make the plugin work properly, the following operations must be authorized for Topic and Cluster resource types:"

doc/docs/modules/ROOT/pages/kafka-ssl.adoc
```diff
@@ -195,5 +195,79 @@ kafka.sasl.mechanism=PLAIN
...
+
+* **Write**, when you want to use the plugin as a Source
+* **Read**, when you want to use the plugin as a Sink
+* **DescribeConfigs** and **Describe**, because the plugin uses the following 2 Kafka AdminClient API:
+    ** listTopics
+    ** describeCluster
+
+About the Streams Procedures you don't need to authorize a particular operation because the producer and the consumer created for the Source and the Sink will be reused by the procedures. This means that if you, for example, authorize the WRITE operation on a topic then you can use both the Source and the Procedure, depending on what you prefer.
```
moxious, line 269: Revise:   "To use streams procedures, the same operations must be authorized (read or write) depending on which of the procedures you wish to use.  The permissions required by the procedures and the source/sink operations are the same."

moxious (changes requested):
Field engineer indicated the PR was good but suggested copy editing for English & grammar readability.   When this review is addressed, it's good to merge.  Thanks for your continued work on this, it was a tough one.

mroiter-larus:
@moxious PR updated as requested

conker84:
thank you @mroiter-larus and @moxious

PR #473, opened 2021-06-07
fixes #472: Fix race condition on startup 4.2.x

### LINKED ISSUE #472: Fix race condition on startup 4.2.x
With Neo4j 4.2.x (tested with 4.2.5 and 4.2.3) I found a possible race condition that prevents the Causal Cluster to form correctly.

conker84:
Fixes #472


conker84:
Meriging this as the Java CI is green

PR #551, opened 2022-12-06
Add kafka-connect sink connector dropUnwind property

### LINKED ISSUE #550: Add Kafka-connect sink connector 'neo4j.cypher.dropUnwind' property
Sink connector (Kafka-connect) does not allow to skip UNWIND prefix from cypher template strategy queries. We are using Kconnect cluster in our migrations tooling and do not always want to process events. Often it is a query without unwinded events  but the lack of this feature sets certain limitations to the queries we can execute means there is no way to execute DML queries such as index of constraint creation and CALLs without YIELD+RETURN.
A new `neo4j.cypher.dropUnwind` property should allow to do that.

igorvoltaic:
Fixes  #550

A new `neo4j.cypher.dropUnwind` property should allow to skip UNWIND prefix from cypher template strategy queries.

Add `neo4j.cypher.dropUnwind` sink connector property


igorvoltaic:
close in favour of 5.0 branch

PR #556, opened 2023-02-19
Fixes #555 Structuring nested object causes DataException

### LINKED ISSUE #555: Structuring nested object causes org.apache.kafka.connect.errors.DataException

Can be structured without exception.

```
org.apache.kafka.connect.errors.ConnectException: org.apache.kafka.connect.errors.DataException: Invalid Java object for schema type MAP: class org.apache.kafka.connect.data.Struct ...
```

Specify a `neo4j.source.query` similar to the following:

```
MATCH (n: Test)
RETURN
     ... (skip) ...
    {
        key1: {
            subKey1: "value",
            subKey2: "value"
        },
        key2: {
            subKey1: "value",
            subKey2: true
        }
    } AS nested
```

## Specifications (Mandatory)

Currently used versions

### Versions

  - Neo4j Streams: 5.0.2

ogawa-takeshi:
Fixes #555

### DIFF
--- a/kafka-connect-neo4j/src/test/kotlin/streams/kafka/connect/source/Neo4jSourceTaskTest.kt
+++ b/kafka-connect-neo4j/src/test/kotlin/streams/kafka/connect/source/Neo4jSourceTaskTest.kt
@@ -15,7 +15,6 @@ import org.junit.Test
 import org.mockito.Mockito
 import org.neo4j.driver.Driver
 import org.neo4j.driver.Session
-import org.neo4j.driver.types.Node
 import org.neo4j.function.ThrowingSupplier
 import streams.Assert
 import streams.Neo4jContainerExtension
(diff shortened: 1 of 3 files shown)


conker84:
Hi @ogawa-takeshi thanks for the contribution but I think there is a misunderstanding. 
A map chan have only values of the same type.
Something like this:

```
{
                |       key1: {
                |           subKey1: "value",
                |           subKey2: "value"
                |       },
                |       key2: {
                |           subKey1: "value",
                |           subKey2: true
                |       }
                |   } AS nested,
```

kafka-connect-neo4j/src/test/kotlin/streams/kafka/connect/source/Neo4jSourceTaskTest.kt
```diff
@@ -274,6 +283,16 @@ class Neo4jSourceTaskTest {
...
+                |   {
+                |       key1: {
+                |           subKey1: "value",
+                |           subKey2: "value"
+                |       },
+                |       key2: {
+                |           subKey1: "value",
+                |           subKey2: true
+                |       }
+                |   } AS nested,
```
conker84, line 295: I think there is a misunderstanding a `Map` can have only values of the same simple type, like for instance`Map<String, Integer>` or `Map<String, String>`.
The structure that you are expressing here is not a `Map`, but a `Struct`.

conker84 (commented):
Hi @ogawa-takeshi thanks for the contribution, please see my comment

conker84:
I'm closing this for lack of feedback. Feel free to re-open

PR #573, opened 2023-07-26
Issue #549: Upgrade kafka-avro-serializer dependency

### LINKED ISSUE #549: Upgrade kafka-avro-serializer dependency
## Feature description 

From Zendesk ticket [#21480](https://neotechnology.zendesk.com/agent/tickets/21480) has been highlighted the need to upgrade `kafka-avro-serializer` dependency from the version 5.2.2 (which includes an old version of Netty library - 3.10.6.Final - with a security vulnerability) to the 7.2.2. 

Could this result into the needs to update also the `kafka-clients` library accordingly? Actually we're using 2.4.1, which has some vulnerabilities (see more details at the following link: https://mvnrepository.com/artifact/org.apache.kafka/kafka-clients/2.4.1)

mroiter-larus:
Fixes #549 

Updated versions of the following libraries:

- io.confluent.kafka.avro.serializer, from 5.2.2 to 7.4.0
- org.apache.avro.avro, from 1.8.2 to 1.11.2
- com.google.guava.guava, from 27.0.1-jre to 32.1.1-jre
- org.apache.kafka.kafka-clients and org.apache.kafka.connect-api, from 2.4.1 to 3.5.1

Fixed also a couple of compilation error in CommonExtension.kt and KafkaErrorServiceTest.kt

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -55,15 +55,15 @@
         <kotlin.version>1.6.10</kotlin.version>
         <kotlin.coroutines.version>1.6.0</kotlin.coroutines.version>
         <neo4j.version>4.4.3</neo4j.version>
-        <kafka.version>2.4.1</kafka.version>
+        <kafka.version>3.5.1</kafka.version>
         <jackson.version>2.13.1</jackson.version>
         <kotlin.compiler.incremental>true</kotlin.compiler.incremental>
         <neo4j.java.driver.version>4.4.2</neo4j.java.driver.version>
         <testcontainers.version>1.15.1</testcontainers.version>
-        <avro.version>1.8.2</avro.version>
+        <avro.version>1.11.2</avro.version>
         <mokito.version>3.3.0</mokito.version>
         <junit.version>4.13.2</junit.version>
-        <kafka.avro.serializer.version>5.2.2</kafka.avro.serializer.version>
+        <kafka.avro.serializer.version>7.4.0</kafka.avro.serializer.version>
         <junit-jupiter.version>5.7.1</junit-jupiter.version>
         <hamcrest.version>1.3</hamcrest.version>
         <neo4j.configuration-lifecycle.version>ad59084711</neo4j.configuration-lifecycle.version>
--- a/pom.xml
+++ b/pom.xml
@@ -56,7 +56,7 @@
         <kotlin.coroutines.version>1.6.0</kotlin.coroutines.version>
         <neo4j.version>4.4.25</neo4j.version>
         <kafka.version>2.6.3</kafka.version>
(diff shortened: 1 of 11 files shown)


pom.xml
```diff
@@ -55,15 +55,15 @@
         <kotlin.version>1.6.10</kotlin.version>
         <kotlin.coroutines.version>1.6.0</kotlin.coroutines.version>
         <neo4j.version>4.4.3</neo4j.version>
-        <kafka.version>2.4.1</kafka.version>
+        <kafka.version>3.5.1</kafka.version>
```
ali-ince, line 58: Let's stick with 2.6.3 as we discussed before with @conker84.

pom.xml
```diff
@@ -55,15 +55,15 @@
         <kotlin.version>1.6.10</kotlin.version>
         <kotlin.coroutines.version>1.6.0</kotlin.coroutines.version>
         <neo4j.version>4.4.3</neo4j.version>
```
ali-ince, line 57: let's upgrade this to the latest also?

pom.xml
```diff
@@ -55,15 +55,15 @@
...
-        <kafka.version>2.4.1</kafka.version>
+        <kafka.version>3.5.1</kafka.version>
         <jackson.version>2.13.1</jackson.version>
```
ali-ince, line 59: this can also be upgraded to 2.15.3.

pom.xml
```diff
@@ -55,15 +55,15 @@
...
         <kotlin.compiler.incremental>true</kotlin.compiler.incremental>
         <neo4j.java.driver.version>4.4.2</neo4j.java.driver.version>
```
ali-ince, line 61: There's 4.4.12

ali-ince (changes requested):
Let's not move to the latest kafka. I also suggested to upgrade other dependencies too, but happy to see it as a separate PR if you prefer so.

pom.xml
conker84: We can skip this is used by the kafka connector which is not used here anymore

pom.xml
conker84: done

pom.xml
conker84: done

conker84:
@ali-ince we had to downgrade to 2.14.3 as with 2.15 was complaining about classes compiled with Java 17.

ali-ince:
also just checked the jackson jdk compatibility and it looks like it is still on java 8? 😕

conker84:
> also just checked the jackson jdk compatibility and it looks like it is still on java 8? 😕

Yep I checked it to, that's odd...

PR #578, opened 2023-08-07
Ids in relationship

### LINKED ISSUE #577: CUD Strategy - Ids in Relations
On CUD strategy allow to pass ids as for relationship, same as for nodes.
It is important for cases that you have multiple relations of the same type but with different properties and ids.
For example (n:Employees)-[ROLE: {uuid: 'some uuid'}-(t:Teams)
In the current implementation there is no way to  merge or delete to a specific role.

It will allow to delete, update a specific relation with the same name between two nodes.
at the moment it can not be specified.

aviz:
Fixes #577

  - Added support to pass ids where merging and deleting a relationship.

### DIFF
--- a/common/src/main/kotlin/streams/extensions/Quadruple.kt
+++ b/common/src/main/kotlin/streams/extensions/Quadruple.kt
@@ -1,0 +1,12 @@
+package streams.extensions
+
+import java.io.Serializable
+
+data class Quadruple<out A, out B, out C, out D>(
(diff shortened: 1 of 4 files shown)


conker84:
@aviz sorry but I don't understand the goal here. There are no test cases and the description is quite skimpy.

conker84:
@aviz I'm closing this for lack of feedback. Feel free to reopen

PR #579, opened 2023-08-10
Fixed #535: Kafka connector Neo4jSourceTaskTest.'should throw exception' is flaky

### LINKED ISSUE #535: Kafka connector Neo4jSourceTaskTest.'should throw exception' is flaky
## Guidelines

Please note that GitHub issues are only meant for bug reports/feature requests. If you have questions on how to use Streams, please ask on [StackOverflow](https://stackoverflow.com/questions/tagged/neo4j-streams) instead of creating an issue here.

The `Neo4jSourceTaskTest.'should throw exception'` should pass.

The test fails sometimes with
`java.lang.AssertionError: Expected exception: org.apache.kafka.connect.errors.ConnectException`.
The test seems to be dependent on the test execution order. More specifically it fails if it runs after `Neo4jSinkTaskTest`.

Run `Neo4jSinkTaskTest` and then `Neo4jSourceTaskTest.'should throw exception'` in the same test run.

### Simple Dataset (where it's possibile)

```
//Insert here a set of Cypher statements that helps us to reproduce the problem

```

### Steps (Mandatory)

  1.
  1.
  1.

## Specifications (Mandatory)

Currently used versions

### Versions

  - OS:
  - Neo4j:
  - Neo4j-Streams:

mroiter-larus:
Fixes #535

### DIFF
--- a/kafka-connect-neo4j/src/test/kotlin/streams/kafka/connect/source/Neo4jSourceTaskTest.kt
+++ b/kafka-connect-neo4j/src/test/kotlin/streams/kafka/connect/source/Neo4jSourceTaskTest.kt
@@ -290,7 +290,17 @@ class Neo4jSourceTaskTest {
         val totalRecords = 10
         insertRecords(totalRecords)
 
-        task.poll()
+        var exception: ConnectException? = null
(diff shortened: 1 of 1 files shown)


conker84:
Thank you @mroiter-larus !

PR #583, opened 2023-08-22
fixes #567: ( Impact on BC/DR - Data loss ) - High priority, Nodes with PointValue property cannot be successfully sink to Neo4j With Neo4j Streams Plugin

### LINKED ISSUE #467: Include in documentation link to repo for quickstart of confluent connector / Aura
See this repo:

https://github.com/alexwoolford/ccloud-neo4j

This should be pointed to in the docs as a quick way to get started with Sink:  Kafka -> Aura.

### LINKED ISSUE #567: ( Impact on BC/DR - Data loss ) - High priority, Nodes with PointValue property cannot be successfully sink to Neo4j With Neo4j Streams Plugin
## Guidelines

Since the original issue (#559) was closed, and the answerer did not understand the issue correctly, so I create a new issue to track.

Please refer to #559 for details.

### Versions

  - Neo4j: 4.3.3
  - Neo4j-Streams: 4.1

conker84:
Fixes #567

  - fixed the bug
  - added test
  - removed kafka-connect from the build as we rely on branch `5.0` for that


common/src/main/kotlin/streams/utils/JSONUtils.kt
```diff
@@ -33,6 +54,42 @@ fun Point.toStreamsPoint(): StreamsPoint {
     }
 }
 
+fun Map<String, Any>.toMapValue(): MapValue {
```
ali-ince, line 57: should we maybe use `org.neo4j.values.virtual.MapValueBuilder` instead?

common/src/main/kotlin/streams/utils/JSONUtils.kt
```diff
@@ -121,6 +178,100 @@ class DriverRelationshipSerializer : JsonSerializer<org.neo4j.driver.types.Relat
...
+    abstract fun convertPayload(payloadMap: JsonNode): PAYLOAD
+    abstract fun fillPayload(payload: PAYLOAD,
+                             beforeProps: Map<String, Any>?,
+                             afterProps: Map<String, Any>?): PAYLOAD
+
+    @Throws(IOException::class, JsonProcessingException::class)
+    override fun deserialize(parser: JsonParser, context: DeserializationContext): EVENT {
+        val root: JsonNode = parser.codec.readTree(parser)
+        val meta = JSONUtils.convertValue<Meta>(root["meta"])
+        val schema = JSONUtils.convertValue<Schema>(root["schema"])
```
ali-ince, line 249: am I correct to think that if the user did not include schema information, this still won't work? If so, what would the user do to create point values from messages generated by non-neo4j applications? Regardless, I think we need to extend the documentation accordingly.

pom.xml
```diff
@@ -43,7 +43,7 @@
     <modules>
         <module>common</module>
         <module>test-support</module>
-        <module>kafka-connect-neo4j</module>
+<!--        <module>kafka-connect-neo4j</module>-->
```
ali-ince, line 46: should we remove this and even the source folder, maybe as a separate PR?

ali-ince (commented):
How about AVRO support? Is it supported out of the box, or do we need to implement that too?

pom.xml
conker84: Yes I commented this to open a discussion about it. As we don't need it anymore we can get rid of it in this branch

common/src/main/kotlin/streams/utils/JSONUtils.kt
conker84: I would not advocate to the users to use this message created by themselves otherwise we need to document all the possible permutations of schema and so on. So for me, this is about the CDC message which contains the info. If it doesn't it does mean that the message is not well-formed.
Maybe is worth adding docs or at least a link to the CDC doc page (https://neo4j.com/docs/kafka-streams/producer/#_transaction_event_handler) in the Kafka Connect PR where there is no direct reference to the message structure.

common/src/main/kotlin/streams/utils/JSONUtils.kt
conker84: done

conker84:
> How about AVRO support? Is it supported out of the box, or do we need to implement that too?

The Streams Transaction Event Handler does not produce messages as AVRO format.

common/src/main/kotlin/streams/utils/JSONUtils.kt
ali-ince: Yes, looks like we are also missing the type for `geo` properties on the sample messages on the page you linked. Would you please update that page and also a separate page (probably best to avoid links) that includes similar sample messages for the kafka connect documentation?

ali-ince (approved):

PR #624, opened 2024-05-10
Fix: Point2D serialization and deserialization

Emrehzl94:
Fixes #<Replace with the number of the issue, Mandatory>

  -
  -
  -

### DIFF
--- a/common/src/main/kotlin/streams/utils/JSONUtils.kt
+++ b/common/src/main/kotlin/streams/utils/JSONUtils.kt
@@ -55,11 +55,17 @@ fun Point.toStreamsPoint(): StreamsPoint {
     }
 }
 
-fun Map<String, Any>.toMapValue(): MapValue {
-    val map = this
-    val builder = MapValueBuilder()
-    map.forEach { (t, u) -> builder.add(t, Values.of(u)) }
-    return builder.build()
+fun Map<String, Any?>.toMapValue(): MapValue {
+    return toMapValue(true)
+}
+
+fun Map<String, Any?>.toMapValue(allowNulls: Boolean): MapValue {
+    return this
+        .filterValues { v -> allowNulls || v != null }
+        .entries
+        .fold(MapValueBuilder()) { builder, entry ->
+            builder.add(entry.key, Values.of(entry.value, allowNulls)); builder}
+        .build()
 }
 
 fun PointValue.toStreamsPoint(): StreamsPoint {
@@ -236,7 +242,7 @@ abstract class StreamsTransactionEventDeserializer<EVENT, PAYLOAD: Payload> : Js
         ?.properties
         ?.mapValues {
             if (points.contains(it.key)) {
-                org.neo4j.values.storable.PointValue.fromMap((it.value as Map<String, Any>).toMapValue().filter { _, u -> u != Values.NO_VALUE })
+                org.neo4j.values.storable.PointValue.fromMap((it.value as Map<String, Any>).toMapValue(false))
             } else {
                 it.value
             }
--- a/common/src/main/kotlin/streams/utils/JSONUtils.kt
+++ b/common/src/main/kotlin/streams/utils/JSONUtils.kt
@@ -17,9 +17,7 @@ import com.fasterxml.jackson.module.kotlin.convertValue
 import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
 import com.fasterxml.jackson.module.kotlin.readValue
 import org.neo4j.driver.internal.value.PointValue
-import org.neo4j.function.ThrowingBiConsumer
 import org.neo4j.graphdb.spatial.Point
-import org.neo4j.values.AnyValue
 import org.neo4j.values.storable.CoordinateReferenceSystem
 import org.neo4j.values.storable.Values
 import org.neo4j.values.virtual.MapValue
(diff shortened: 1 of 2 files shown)


common/src/test/kotlin/streams/utils/JSONUtilsTest.kt
```diff
@@ -111,6 +111,81 @@ class JSONUtilsTest {
...
+        val cdcMap = mapOf<String, Any>(
+            "meta" to mapOf("timestamp" to timestamp,
+                "username" to "user",
+                "txId" to 1,
+                "txEventId" to 0,
+                "txEventsCount" to 1,
+                "operation" to OperationType.created),
+            "payload" to mapOf("id" to "0",
+                "before" to null,
+                "after" to NodeChange(properties = mapOf("location" to mapOf("crs" to "wgs-84", "latitude" to 12.78, "longitude" to 56.7, "height" to null)),
```
fbiville, line 130: is the "height" attribute required here? can we skip it?

common/src/test/kotlin/streams/utils/JSONUtilsTest.kt
```diff
@@ -111,6 +111,81 @@ class JSONUtilsTest {
...
+                "operation" to OperationType.created),
+            "payload" to mapOf("id" to "0",
+                "before" to null,
+                "after" to NodeChange(properties = mapOf("location" to mapOf("crs" to "cartesian", "x" to 12.78, "y" to 56.7, "z" to null)),
+                    labels = listOf("LabelCDC")),
+                "type" to EntityType.node),
+            "schema" to mapOf("properties" to mapOf("location" to "PointValue"))
+        )
+
+        val cdcString = """{
```
fbiville, line 173: this (and its related assertions) can be moved to a separate test imo

common/src/main/kotlin/streams/utils/JSONUtils.kt
```diff
@@ -236,7 +236,7 @@ abstract class StreamsTransactionEventDeserializer<EVENT, PAYLOAD: Payload> : Js
         ?.properties
         ?.mapValues {
             if (points.contains(it.key)) {
-                org.neo4j.values.storable.PointValue.fromMap((it.value as Map<String, Any>).toMapValue())
+                org.neo4j.values.storable.PointValue.fromMap((it.value as Map<String, Any>).toMapValue().filter { _, u -> u != Values.NO_VALUE })
```
fbiville, line 239: you could instead update `streams.utils.JSONUtilsKt#toMapValue` to something like this:

```kotlin
fun Map<String, Any?>.toMapValue(): MapValue {
    return this
        .filterValues { v -> v != null }
        .entries
        .fold(MapValueBuilder()) { builder, entry -> 
            builder.add(entry.key, Values.of(entry.value, false)); builder}
        .build()
}
```

I even wonder if we need an extension function like this at all... `toMapValue` is only used in 2 tests and here. A regular function (or no function at all) would maybe work as well. 
Please lmk what you think :)

common/src/test/kotlin/streams/utils/JSONUtilsTest.kt
Emrehzl94: It's actually part of the test case to see if we're able to deserialize it without having any issues. The reason I did it this way is even though we prevented serializer from generating null values, there might still be some old events generating height = null value, so we decided to cover this case as well.

common/src/test/kotlin/streams/utils/JSONUtilsTest.kt
Emrehzl94: yes, this was also another idea, I just followed the pattern of the other unit tests in this class but of course separating them much cleaner I think.

common/src/main/kotlin/streams/utils/JSONUtils.kt
Emrehzl94: yes, this also crossed my mind but I think this is a general method that can also be used in other places to map values. This is why I decided to add the NO_VALUE filter in the point specific method. But of course getting rid of toMapValue method could also work.

common/src/test/kotlin/streams/utils/JSONUtilsTest.kt
fbiville: let's split this test into 2 then, if that's ok with you?

common/src/main/kotlin/streams/utils/JSONUtils.kt
fbiville: maybe this then?

```kotlin
fun Map<String, Any?>.toMapValue(allowNulls: boolean): MapValue {
    return this
        .filterValues { v -> allowNulls || v != null }
        .entries
        .fold(MapValueBuilder()) { builder, entry -> 
            builder.add(entry.key, Values.of(entry.value, allowNulls)); builder}
        .build()
}
```

common/src/test/kotlin/streams/utils/JSONUtilsTest.kt
fbiville: ah yes, sorry, got confused about the inputs and outputs...
I think splitting the test in 2 as suggested below and a couple of renames and/or explicit `//when`/`//then` comments would help here

fbiville (changes requested):

common/src/test/kotlin/streams/utils/JSONUtilsTest.kt
Emrehzl94: 👍

common/src/main/kotlin/streams/utils/JSONUtils.kt
Emrehzl94: yes, I think this works

fbiville (approved):

PR #631, opened 2024-06-28
fix: latitude and longitude mapping issue

yhhongyang:
Fixes #<Replace with the number of the issue, Mandatory>

According to neo4j [PointValue document](https://neo4j.com/docs/cypher-manual/4.4/values-and-types/spatial/#spatial-values-spatial-instants), x represent longitude and y represent latitude  

Change the mapping of latitude and longitude

  -
  -
  -


fbiville:
CLA resolution in progress in https://github.com/neo4j-contrib/neo4j-streams/pull/632.

PR #632, opened 2024-06-28
fix: latitude and longitude mapping issue for version 5.0

yhhongyang:
According to neo4j [PointValue document](https://neo4j.com/docs/cypher-manual/4.4/values-and-types/spatial/#spatial-values-spatial-instants), x represent longitude and y represent latitude

Change the mapping of latitude and longitude

PR #634, opened 2024-07-16
fix: longitude and latitude mapping 5.0

Emrehzl94:
This pr fixes point value mapping issue for version 5.0


ali-ince:
Duplicate of #632.

PR #642, opened 2024-07-26
feat: Add connector configuration migrator for 5.1

dhrudevalia:
## Issue

On upgrading the connector to 5.1, there are breaking configuration changes.

## Proposed Changes

On connector shutdown, a migrated configuration using new 5.1 configuration keys are dumped to the log to be used after connector migration.

### DIFF
--- a/kafka-connect-neo4j/src/main/kotlin/streams/kafka/connect/common/ConfigurationMigrator.kt
+++ b/kafka-connect-neo4j/src/main/kotlin/streams/kafka/connect/common/ConfigurationMigrator.kt
@@ -1,0 +1,152 @@
+package streams.kafka.connect.common
+
+import com.fasterxml.jackson.databind.ObjectMapper
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+import streams.kafka.connect.common.Neo4jConnectorConfig.Companion.AUTHENTICATION_BASIC_PASSWORD
+import streams.kafka.connect.common.Neo4jConnectorConfig.Companion.AUTHENTICATION_BASIC_REALM
+import streams.kafka.connect.common.Neo4jConnectorConfig.Companion.AUTHENTICATION_BASIC_USERNAME
+import streams.kafka.connect.common.Neo4jConnectorConfig.Companion.AUTHENTICATION_KERBEROS_TICKET
+import streams.kafka.connect.common.Neo4jConnectorConfig.Companion.AUTHENTICATION_TYPE
+import streams.kafka.connect.common.Neo4jConnectorConfig.Companion.BATCH_SIZE
+import streams.kafka.connect.common.Neo4jConnectorConfig.Companion.BATCH_TIMEOUT_MSECS
+import streams.kafka.connect.common.Neo4jConnectorConfig.Companion.CONNECTION_LIVENESS_CHECK_TIMEOUT_MSECS
+import streams.kafka.connect.common.Neo4jConnectorConfig.Companion.CONNECTION_MAX_CONNECTION_ACQUISITION_TIMEOUT_MSECS
+import streams.kafka.connect.common.Neo4jConnectorConfig.Companion.CONNECTION_MAX_CONNECTION_LIFETIME_MSECS
+import streams.kafka.connect.common.Neo4jConnectorConfig.Companion.CONNECTION_POOL_MAX_SIZE
+import streams.kafka.connect.common.Neo4jConnectorConfig.Companion.DATABASE
+import streams.kafka.connect.common.Neo4jConnectorConfig.Companion.ENCRYPTION_CA_CERTIFICATE_PATH
+import streams.kafka.connect.common.Neo4jConnectorConfig.Companion.ENCRYPTION_ENABLED
+import streams.kafka.connect.common.Neo4jConnectorConfig.Companion.ENCRYPTION_TRUST_STRATEGY
+import streams.kafka.connect.common.Neo4jConnectorConfig.Companion.RETRY_BACKOFF_MSECS
+import streams.kafka.connect.common.Neo4jConnectorConfig.Companion.RETRY_MAX_ATTEMPTS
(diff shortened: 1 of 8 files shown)


kafka-connect-neo4j/src/main/kotlin/streams/kafka/connect/common/ConfigurationMigrator.kt
```diff
@@ -0,0 +1,164 @@
...
+                        log.debug("Migrating configuration prefix key {} to {}", originalKey, newKey)
+                    }
+                }
+            } else {
+                // Configuration option not declared should be copied across
+                updatedConfig[originalKey] = value
+            }
+        }
+
+        outputToLog(updatedConfig)
```
venikkin, line 138: This line is here is very important. At the same time, neither the returned type nor the method name doesn't reflect that we print the config into the log. 
I suggest removing this line and writing a separate (maybe utility) method (something like `printAsJson(o: Any)`), and passing the updated/migrated config to this method. I think think might be cleaner. 
Maybe it could be a 3-liner instead:
```
val migrated = migrator.migrate()
val json = mapper.writeValueAsString(migrated)
log.info(...)
```
which is also IMO slightly better.

venikkin (approved):
Look good. Small suggestion with the `migrate` method.

kafka-connect-neo4j/src/main/kotlin/streams/kafka/connect/common/ConfigurationMigrator.kt
```diff
@@ -0,0 +1,164 @@
...
+        Neo4jSinkConnectorConfig.TOPIC_CYPHER_PREFIX to "neo4j.cypher.topic."
+    )
+
+    /**
+     * Migrate configuration keys from existing format to v5.1 connector format.
+     * Configuration properties containing msecs units are converted to new format.
+     *
+     * @return updated configuration key-value pairs
+     */
+    fun migrate(): Map<String, Any> {
```
fbiville, line 112: nit: I'd go as far as explicitly naming this "migrateToV51"

kafka-connect-neo4j/src/test/kotlin/streams/kafka/connect/common/ConfigurationMigratorTest.kt
```diff
@@ -0,0 +1,178 @@
...
+      mapOf(
+        "neo4j.topic.pattern.merge.node.properties.enabled" to "true",
+        "neo4j.server.uri" to "neo4j+s://x.x.x.x",
+        "neo4j.retry.max.attemps" to "1"
+      )
+
+    // When the configuration is migrated
+    val migratedConfig = ConfigurationMigrator(originals).migrate()
+
+    // Then the keys are updated to new key format containing the original value
```
fbiville, line 26: nit: assert the size of `migratedConfig` as well?

fbiville (approved):
Very nice, thanks!

venikkin:
[Dump a migrated configuration for the old connector during shutdown](https://trello.com/c/HN850BYF/576-dump-a-migrated-configuration-for-the-old-connector-during-shutdown)

kafka-connect-neo4j/src/main/kotlin/streams/kafka/connect/common/ConfigurationMigrator.kt
dhrudevalia: Is there a clearer way to represent 5.1 in camelcase?
V51 to me looks like it could be version 51.
e.g. 5_1, 5dot1?

kafka-connect-neo4j/src/main/kotlin/streams/kafka/connect/common/ConfigurationMigrator.kt
dhrudevalia: Just clarifying that you're suggesting uncoupling the logging from this class and instead doing it in, `SinkConnect/SourceConnector::stop` instead?

kafka-connect-neo4j/src/test/kotlin/streams/kafka/connect/common/ConfigurationMigratorTest.kt
dhrudevalia: done

kafka-connect-neo4j/src/main/kotlin/streams/kafka/connect/source/Neo4jSourceConnector.kt
```diff
@@ -27,7 +25,9 @@ class Neo4jSourceConnector: SourceConnector() {
         config = Neo4jSourceConnectorConfig(settings)
     }
 
-    override fun stop() {}
+    override fun stop() {
+        ConfigurationMigrator(settings).migrate()
```
ali-ince, line 29: Should we also include stored offset value (by dumping the configuration by setting `start-from` to `user defined` and `start-from.value` to the last stored offset) with the source connector, wdyt?

dhrudevalia: Is this a configuration option that didn't exist in the original connector, but does in 5.1? Would that be handled by defaults?

kafka-connect-neo4j/src/main/kotlin/streams/kafka/connect/common/ConfigurationMigrator.kt
fbiville: I'll let you pick and/or let others chime in, I have no strong preference here

kafka-connect-neo4j/src/main/kotlin/streams/kafka/connect/common/ConfigurationMigrator.kt
dhrudevalia: done: 3-liner approach

fbiville (approved):

PR #651, opened 2024-09-04
fix: obfuscate password in configuration migration log

dhrudevalia:
Fixes printing password to log during configurtion migration

  - Do not migrate `neo4j.authentication.basic.password` property value. Maintain the key but replace the value with an empty string

### DIFF
--- a/kafka-connect-neo4j/src/main/kotlin/streams/kafka/connect/common/ConfigurationMigrator.kt
+++ b/kafka-connect-neo4j/src/main/kotlin/streams/kafka/connect/common/ConfigurationMigrator.kt
@@ -66,7 +66,7 @@ class ConfigurationMigrator(private val settings: Map<String, String>) {
         AUTHENTICATION_BASIC_USERNAME to PropertyConverter("neo4j.authentication.basic.username") {settings[AUTHENTICATION_BASIC_USERNAME] as String},
         AUTHENTICATION_BASIC_PASSWORD to PropertyConverter("neo4j.authentication.basic.password") {""},
         AUTHENTICATION_BASIC_REALM to PropertyConverter("neo4j.authentication.basic.realm") {settings[AUTHENTICATION_BASIC_REALM] as String},
-        AUTHENTICATION_KERBEROS_TICKET to PropertyConverter("neo4j.authentication.kerberos.ticket") {settings[AUTHENTICATION_KERBEROS_TICKET] as String},
+        AUTHENTICATION_KERBEROS_TICKET to PropertyConverter("neo4j.authentication.kerberos.ticket") {""},
         BATCH_SIZE to PropertyConverter("neo4j.batch-size") {settings[BATCH_SIZE] as String},
         ENCRYPTION_ENABLED to PropertyConverter("neo4j.security.encrypted") {settings[ENCRYPTION_ENABLED] as String},
         ENCRYPTION_TRUST_STRATEGY to PropertyConverter("neo4j.security.trust-strategy") {settings[ENCRYPTION_TRUST_STRATEGY] as String},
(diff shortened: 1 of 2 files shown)


kafka-connect-neo4j/src/main/kotlin/streams/kafka/connect/common/ConfigurationMigrator.kt
```diff
@@ -64,7 +64,7 @@ class ConfigurationMigrator(private val settings: Map<String, String>) {
         SERVER_URI to PropertyConverter("neo4j.uri") { settings[SERVER_URI] as String },
         AUTHENTICATION_TYPE to PropertyConverter("neo4j.authentication.type") { settings[AUTHENTICATION_TYPE] as String },
         AUTHENTICATION_BASIC_USERNAME to PropertyConverter("neo4j.authentication.basic.username") {settings[AUTHENTICATION_BASIC_USERNAME] as String},
-        AUTHENTICATION_BASIC_PASSWORD to PropertyConverter("neo4j.authentication.basic.password") {settings[AUTHENTICATION_BASIC_PASSWORD] as String},
+        AUTHENTICATION_BASIC_PASSWORD to PropertyConverter("neo4j.authentication.basic.password") {""},
```
ali-ince, line 67: maybe `***`?

kafka-connect-neo4j/src/main/kotlin/streams/kafka/connect/common/ConfigurationMigrator.kt
```diff
@@ -64,7 +64,7 @@ class ConfigurationMigrator(private val settings: Map<String, String>) {
...
         AUTHENTICATION_BASIC_REALM to PropertyConverter("neo4j.authentication.basic.realm") {settings[AUTHENTICATION_BASIC_REALM] as String},
         AUTHENTICATION_KERBEROS_TICKET to PropertyConverter("neo4j.authentication.kerberos.ticket") {settings[AUTHENTICATION_KERBEROS_TICKET] as String},
```
ali-ince, line 69: I think we should do the same for this one too.

kafka-connect-neo4j/src/main/kotlin/streams/kafka/connect/common/ConfigurationMigrator.kt
dhrudevalia: I originally thought `***` as well, but i'm assuming that the error message we would get for an empty password rather than incorrect password would be more informative to the user that it hasn't been filled.
I'm hoping it would be something like 'No password found' vs  'invalid password'

kafka-connect-neo4j/src/main/kotlin/streams/kafka/connect/common/ConfigurationMigrator.kt
dhrudevalia: Sorry not sure which line is being highlighted. Is it realm + kerberos ticket?

ali-ince: the kerberos ticket one.

kafka-connect-neo4j/src/main/kotlin/streams/kafka/connect/common/ConfigurationMigrator.kt
ali-ince: ok, that also makes sense. let's keep it as empty as you suggested.

kafka-connect-neo4j/src/main/kotlin/streams/kafka/connect/common/ConfigurationMigrator.kt
dhrudevalia: added

ali-ince (approved):

PR #724, opened 2025-02-17
feat: switch to caniuse and add cypher version selection

ali-ince:
This PR switches to caniuse library and starts prefixing generated cypher statements with `CYPHER 5` on supported versions.


ali-ince:
closing in favour of #725 .

